PR #2, opened 2022-07-06
Docusaurus PR for K3s documentation

divya-mohan0209:
Docusaurus PR for K3s documentation


brandond:
This is fantastic, thanks for all the hard work in converting the site over!

My only concern at the moment is that the yellow from the k3s logo is a bit too light for use as a link/highlight color; I find it a bit hard to read. Is there another complimentary color, or perhaps just a darker shade, that we could use instead?

image

dereknola:
image
image

PR #5, opened 2022-07-13
Adding correct URL

divya-mohan0209:
Fixing URL for k3s-io


divya-mohan0209:
cc: @dereknola @cwayne18

PR #25, opened 2022-09-19
update guides

mitulshah-suse:
Related to https://github.com/rancher/rancher-docs/issues/34

### DIFF
--- a/docs/security/self-assessment.md
+++ b/docs/security/self-assessment.md
@@ -464,7 +464,7 @@ journalctl -D /var/log/journal -u k3s | grep 'Running kube-apiserver' | tail -n1
 **Remediation:**
 Follow the documentation and configure alternate mechanisms for authentication. Then,
 edit the API server pod specification file /etc/kubernetes/manifests/kube-apiserver.yaml
-on the control plane node and remove the --token-auth-file=<filename> parameter.
+on the control plane node and remove the `--token-auth-file=<filename>` parameter.
 
 **Audit:**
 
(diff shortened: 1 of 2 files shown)


docs/security/self-assessment.md
```diff
@@ -3,13 +3,13 @@ title: CIS Self Assessment Guide
 weight: 90
 ---
 
-### CIS Kubernetes Benchmark v1.6 - K3s with Kubernetes v1.17 to v1.21
+### CIS Kubernetes Benchmark v1.23 - K3s with Kubernetes v1.22 to v1.24
```
brandond, line 6: We don't have versioned docs, and still need to support 1.22-1.24, so I don't think we can just wholesale replace the 1.6 benchmark docs with content for 1.23.  Perhaps we should move the old content to `self-assessment-1.6.md` and have a `self-assessment-1.23.md` with the new content?

brandond (changes requested):

brandond:
Don't forget to sign your commits for DCO

docs/security/self-assessment.md
mitulshah-suse: K3s support is only for versions 1.22-1.24 in the latest versions of rancher 2.6
Older versions 1.17-1.21 are not supported by K3s. So it should be ok to keep only one version for now.
We might need to add versioning later on though.

mitulshah-suse:
> Don't forget to sign your commits for DCO

Done.

brandond (approved):

PR #28, opened 2022-09-27
Add missing CNAME

dereknola:
Add CNAME for GH pages

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+docs.k3s.io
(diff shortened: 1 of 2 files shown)


cwayne18 (approved):

brandond (approved):

brandond (changes requested):
Actually, I think we need to add a CNAME file. The current URL is fine, we just need GH pages to allow the cname.

brandond (approved):

PR #29, opened 2022-09-27
Move CNAME into GH Actions config

dereknola:
Fix based off https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-add-cname-file-cname


brandond (approved):
🎉

PR #50, opened 2022-11-25
Updated flags for etcd-backup

mattups:
Hello, I updated the flags in the on-demand etcd snapshot snippet to have also the `etcd` prefix. I found after some testing that using `--s3-*` flags doesn't work properly. Fixing the flags to `--etcd-s3` as specified in the docs made the command work.

### DIFF
--- a/docs/backup-restore/backup-restore.md
+++ b/docs/backup-restore/backup-restore.md
@@ -88,11 +88,12 @@ The arguments below have been added to the `server` subcommand. These flags exis
 To perform an on-demand etcd snapshot and save it to S3:
 
 ```bash
+# Specify also --etcd-s3-endpoint if you have a custom endpoints otherwise it would be defaulted to s3.amazonaws.com
 k3s etcd-snapshot \
   --s3 \
-  --s3-bucket=<S3-BUCKET-NAME> \
-  --s3-access-key=<S3-ACCESS-KEY> \
-  --s3-secret-key=<S3-SECRET-KEY>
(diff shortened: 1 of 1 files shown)


dereknola:
Hello @mattups, thanks for your community PR. However, testing with the latest version of K3s, the `etcd-snapshot` tool does work with `--s3*` flags.
```
k3s etcd-snapshot --s3 --s3-endpoint="localhost:9444" --s3-bucket=k3s-test --s3-access-key=AKIAIOSFODNN7EXAMPLE --s3-insecure
...
INFO[0000] Saving etcd snapshot on-demand-degion-1669662500 to S3 
INFO[0000] Checking if S3 bucket k3s-test exists        
INFO[0000] S3 bucket k3s-test exists                    
INFO[0000] Uploading snapshot /var/lib/rancher/k3s/server/db/snapshots/on-demand-degion-1669662500 to S3 
INFO[0000] S3 upload complete for on-demand-degion-1669662500 
INFO[0000] Reconciling etcd snapshot data in k3s-etcd-snapshots ConfigMap 
INFO[0000] Reconciliation of snapshot data in k3s-etcd-snapshots ConfigMap complete 

```

I will note that the `k3s server` does not have the `s3*` flags, but only the `--etcd-s3*` versions. We specifically added aliases to the `etcd-snapshot` command so users could use the `etcd-s3` in their `config.yaml` and it would work for both commands.

mattups:
Hey @dereknola, sorry for the late reply, and thanks for you message!
Forgot to say I was working with `v1.24.8+k3s1`. In any case, thanks for the clarification man!

PR #56, opened 2022-12-04
network-options: fix internal link

vazhnov:
Fix internal link to https://docs.k3s.io/installation/network-options#single-stack-ipv6-installation


brandond:
Thanks for the PR. I think I broke this just recently with some section title changes; we need to add a link checker to the CI scripts.

Issue #41, opened 2022-11-06
Docs might need updating to reflect PSP depreciation


udf2457:
At present the docs do a good job of explaining [how to harden k3s by using PSPs](https://docs.k3s.io/security/hardening-guide).

However, it appears that PSPs are [deprecated as of Kubernetes 1.21](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/)

This likely means that the docs should be updated accordingly.

Unfortunate I'm far from a subject-matter expert in this area, so I can't viably propose a PR myself, hence I am just flagging this here.

cwayne18:
This is a good point and we're aware-- to be fair though they're still usable up until 1.25, so we can't remove the PSP bits while we still support < 1.25, though a fair point we should add PSA docs for 1.25+

PR #59, opened 2022-12-16
Add section on CIS 1.23, PSA in K3s 1.25+ in hardening guide

dereknola:
Addresses https://github.com/k3s-io/docs/issues/41
- Use tabs to rework hardening guide sections
- Section on K3s 1.25+ and the relevant PSA configuration
- Move the K3s "hardened flags" up on the guide, push reference details down. 
- Hardened flags are now provided as a `config.yaml`, the user is already doing prep work on the node, placing a premade config.yaml is inline with other prep work.

### DIFF
--- a/docs/security/hardening-guide.md
+++ b/docs/security/hardening-guide.md
@@ -560,6 +560,165 @@ sudo systemctl daemon-reload
 sudo systemctl restart k3s.service
 ```
 
+## Configuration for Kubernetes Components
+
+
+The configuration below should be placed in the [configuration file](../installation/configuration.md#configuration-file), and contains all the necessary remediations to harden the Kubernetes components.
+
+
+<Tabs>
+<TabItem value="v1.25 and Newer" default>
+
+```yaml
+protect-kernel-defaults: true
+secrets-encryption: true
+kube-apiserver-arg:
+  - 'enable-admission-plugins=NodeRestriction,NamespaceLifecycle,ServiceAccount'
+  - 'admission-control-config-file=/var/lib/rancher/k3s/server/psa.yaml'
+  - 'audit-log-path=/var/lib/rancher/k3s/server/logs/audit.log'
+  - 'audit-policy-file=/var/lib/rancher/k3s/server/audit.yaml'
+  - 'audit-log-maxage=30'
+  - 'audit-log-maxbackup=10'
+  - 'audit-log-maxsize=100'
+  - 'request-timeout=300s'
+  - 'service-account-lookup=true'
+kube-controller-manager-arg:
+  - 'terminated-pod-gc-threshold=10'
+  - 'use-service-account-credentials=true'
+kubelet-arg:
+  - 'streaming-connection-idle-timeout=5m'
+  - 'make-iptables-util-chains=true'
+```
+
+</TabItem>
+
+<TabItem value="v1.24 and Older" default>
+
+```yaml
+protect-kernel-defaults: true
+secrets-encryption: true
+kube-apiserver-arg:
+  - 'enable-admission-plugins=NodeRestriction,PodSecurityPolicy,NamespaceLifecycle,ServiceAccount'
+  - 'audit-log-path=/var/lib/rancher/k3s/server/logs/audit.log'
+  - 'audit-policy-file=/var/lib/rancher/k3s/server/audit.yaml'
+  - 'audit-log-maxage=30'
+  - 'audit-log-maxbackup=10'
+  - 'audit-log-maxsize=100'
+  - 'request-timeout=300s'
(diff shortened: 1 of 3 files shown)


docs/security/hardening-guide.md
```diff
@@ -36,13 +39,57 @@ kernel.keys.root_maxbytes=25000000
...
+The runtime requirements to comply with the CIS Benchmark are centered around pod security (via PSP or PSA), network policies and API Server auditing logs. These are outlined in this section.
+
+By default, K3s does not apply any pod security or network policies. However, K3s ships with a controller that is meant to apply a given set of network policies. The same setup applies API Server auditing logs, K3s doesn't enable them by default, so audit log configuration and audit policy must be created manually. By default, K3s runs with the `NodeRestriction` admission controller.
```
brandond, line 44: What controller do we have that applies network policies?

brandond, line 44: What controller do we have that applies network policies? Do you mean that it enforces the policies, if the user applies (creates) rules in the cluster?

```suggestion
By default, K3s does not include any pod security or network policies. However, K3s ships with a controller that will enforce network policies, if any are created. K3s doesn't enable auditing by default, so audit log configuration and audit policy must be created manually. By default, K3s runs with the both the `PodSecurity` and `NodeRestriction` admission controllers enabled, among others.
```

docs/security/hardening-guide.md
```diff
@@ -464,6 +560,165 @@ sudo systemctl daemon-reload
...
+
+
+<Tabs>
+<TabItem value="v1.25 and Newer" default>
+
+```yaml
+protect-kernel-defaults: true
+secrets-encryption: true
+kube-apiserver-arg:
+  - 'enable-admission-plugins=NodeRestriction,NamespaceLifecycle,ServiceAccount'
```
brandond, line 576: Aren't all three of these these enabled by default?

docs/security/hardening-guide.md
```diff
@@ -464,6 +560,165 @@ sudo systemctl daemon-reload
...
+
+</TabItem>
+
+<TabItem value="v1.24 and Older" default>
+
+```yaml
+protect-kernel-defaults: true
+secrets-encryption: true
+kube-apiserver-arg:
+  - 'enable-admission-plugins=NodeRestriction,PodSecurityPolicy,NamespaceLifecycle,ServiceAccount'
```
brandond, line 601: same question

docs/security/hardening-guide.md
dereknola: GH review is somewhat weird, do you mean: secrets-encryption, protect-kernel-defaults, and the apiserver args as the 3? or the 3 admission plugins?

docs/security/hardening-guide.md
dereknola: My dumb answer is: I don't know, this is an paragraph from the existing documentation. As such I am totally fine with taking your suggestions as fact.

docs/security/hardening-guide.md
dereknola: The answer is no for both. The only default that is enabled for the plugins is `NodeRestriction` so that can be removed from this guide.

dereknola: Which you can't do because once you define the plugins as an argument, it overrides the k3s default (just tested). So all 3 must be included.

brandond: According to the upstream docs:
> --enable-admission-plugins strings
> admission plugins that should be enabled in addition to default enabled ones (NamespaceLifecycle, LimitRanger, ServiceAccount, TaintNodesByCondition, PodSecurity, Priority, DefaultTolerationSeconds, DefaultStorageClass, StorageObjectInUseProtection, PersistentVolumeClaimResize, RuntimeClass, CertificateApproval, CertificateSigning, CertificateSubjectRestriction, DefaultIngressClass, MutatingAdmissionWebhook, ValidatingAdmissionPolicy, ValidatingAdmissionWebhook, ResourceQuota).

So these are **additional** plugins, beyond the defaults - and NamespaceLifecyle and ServiceAccount are already covered by the defaults, so they don't need to be included, either in the K3s defaults, or those set by the user.

dereknola: Doing some testing, I have removed these from the 1.25 config.

docs/security/hardening-guide.md
dereknola: This line cannot be removed from the 1.24 and Older, as older version of Kubernetes such as v1.22 did not enable NodeRestriciton or PodSecurityPolicy by default.

brandond (approved):

rbrtbnfgl (approved):

PR #64, opened 2023-01-05
Add codespell

cwayne18:
so that we can spell gud

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,0 +1,8 @@
+aks
+ec2
+eks
+gce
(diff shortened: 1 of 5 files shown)


briandowns:
You'll need this file: https://github.com/k3s-io/k3s/blob/master/.codespellignore as well given the config added.

cwayne18:
yep just pushed that, thanks @briandowns

brandond (approved):

dereknola (approved):

PR #70, opened 2023-01-11
Document how to configure an HTTP proxy for K3s and Containerd

bashofmann:
Similar documentation already exists for RKE2: https://docs.rke2.io/advanced#configuring-an-http-proxy

### DIFF
--- a/docs/advanced/advanced.md
+++ b/docs/advanced/advanced.md
@@ -57,12 +57,12 @@ The K3s installation script will automatically take the `HTTP_PROXY`, `HTTPS_PRO
 
 Of course, you can also configure the proxy by editing these files.
 
-The `NO_PROXY` variable must include your internal networks, as well as the cluster pod and service IP ranges.
+K3s will automatically add the cluster internal Pod and Service IP ranges and cluster DNS domain to the list of `NO_PROXY` entries. You should ensure that the IP address ranges used by the Kubernetes nodes themselves (i.e. the public and private IPs of the nodes) are included in the `NO_PROXY` list, or that the nodes can be reached through the proxy.
 
 ```
 HTTP_PROXY=http://your-proxy.example.com:8888
 HTTPS_PROXY=http://your-proxy.example.com:8888
-NO_PROXY=127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.svc,.cluster.local
+NO_PROXY=127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
 ```
 
 If you want to configure the proxy settings for containerd without affecting K3s and the Kubelet, you can prefix the variables with `CONTAINERD_`:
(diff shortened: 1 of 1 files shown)


brandond (changes requested):
Unlike RKE2, the K3s installer captures any proxy-related environment variables present when the installer is run, and places them in the environment file managed by the installer - usually `/etc/systemd/system/k3s.env` or `/etc/systemd/system/k3s-agent.env`. For this reason, most users won't need to explicitly perform these steps, as it is done for them. This is probably worth calling out in the docs.

bashofmann:
Good point, I added a note for this.

docs/advanced/advanced.md
```diff
@@ -46,6 +46,33 @@ Any file found in `/var/lib/rancher/k3s/server/manifests` will automatically be
 
 For information about deploying Helm charts, refer to the section about [Helm.](../helm/helm.md)
 
+## Configuring an HTTP proxy
+
+If you are running K3s in an environment, which only has external connectivity through an HTTP proxy, you can configure your proxy settings on the K3s systemd service. These proxy settings will then be used in K3s and passed down to the embedded containerd and kubelet.
+
+The K3s installation script will automatically take the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`, as well as `CONTAINERD_HTTP_PROXY`, `CONTAINERD_HTTPS_PROXY` and `CONTAINERD_NO_PROXY` variables from the current shell, if they are present, and write them to the environment file of your systemd service, usually:
```
brandond, line 53:
```suggestion
The K3s installation script will automatically take the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`, as well as the `CONTAINERD_HTTP_PROXY`, `CONTAINERD_HTTPS_PROXY` and `CONTAINERD_NO_PROXY` variables from the current shell, if they are present, and write them to the environment file of your systemd service, usually:
```

docs/advanced/advanced.md
```diff
@@ -46,6 +46,33 @@ Any file found in `/var/lib/rancher/k3s/server/manifests` will automatically be
...
+
+- `/etc/systemd/system/k3s.service.env`
+- `/etc/systemd/system/k3s-agent.service.env`
+
+Of course, you can also configure the proxy by editing these files.
+
+The `NO_PROXY` variable must include your internal networks, as well as the cluster pod and service IP ranges.
```
brandond, line 60: The RKE2 docs are actually incorrect. The cluster pod and service CIDRs, as well as the cluster domain, are automatically added to the NO_PROXY list by RKE2, and don't need to be added by hand.

The user SHOULD ensure that the address ranges used by the nodes themselves (ie the public/private IPs of the nodes) are included in the NO_PROXY list, or that the nodes can be reached through the proxy.

bashofmann: I'll update it.

Once this is merged, I'll also create a PR for the RKE2 docs then to update it there.

bashofmann: Done.

brandond (approved):

dereknola (approved):

PR #78, opened 2023-01-25
Fix the default egress-selector-mode

PaulSD:
The default `egress-selector-mode` used to be `pod`, but it was changed to `agent` a while back: https://github.com/k3s-io/k3s/commit/0581808f5c160b0c0cafec5b8f20430835f34f44 https://github.com/k3s-io/k3s/blob/master/pkg/cli/cmds/server.go#L229

This commit updates the docs to reflect the current default `egress-selector-mode`.


dereknola (approved):

dereknola:
@PaulSD Nice catch, thanks for the PR!

briandowns (approved):

PR #81, opened 2023-01-31
Fix typos

rjschwei:
Remove use of double "the"
Signed-off-by: Robert Schweikert <rjschwei@suse.com>


cwayne18:
Nice catch, thanks for the PR!

cwayne18 (approved):

PR #95, opened 2023-02-17
Fixed registration address not mentioned

adokay:
Added fixed registration address recommendation.


dereknola:
A fixed registration address is not required to join agents to servers. Agents can pass any `--server <SERVER_IP>` and properly connect to the cluster.

PR #96, opened 2023-02-17
Random fixes for random issues

dereknola:
Addresses:
- https://github.com/k3s-io/k3s/issues/3207
- https://github.com/k3s-io/k3s/issues/2468
- https://github.com/k3s-io/k3s/issues/2059
- https://github.com/k3s-io/k3s/issues/1932
- https://github.com/k3s-io/k3s/issues/1556
- https://github.com/k3s-io/k3s/issues/1309
- https://github.com/k3s-io/k3s/issues/401

### DIFF
--- a/docs/advanced/advanced.md
+++ b/docs/advanced/advanced.md
@@ -149,9 +149,10 @@ sudo etcdctl version \
 
 K3s will generate config.toml for containerd in `/var/lib/rancher/k3s/agent/etc/containerd/config.toml`.
 
-For advanced customization for this file you can create another file called `config.toml.tmpl` in the same directory and it will be used instead.
+For advanced customization for this file you can create another file called `config.toml.tmpl` in the same directory, and it will be used instead.
 
 The `config.toml.tmpl` will be treated as a Go template file, and the `config.Node` structure is being passed to the template. See [this folder](https://github.com/k3s-io/k3s/blob/master/pkg/agent/templates) for Linux and Windows examples on how to use the structure to customize the configuration file.
+The config.Node golang struct is defined [here](https://github.com/k3s-io/k3s/blob/master/pkg/daemons/config/types.go#L37)
 
 ## NVIDIA Container Runtime Support
 
@@ -320,6 +321,19 @@ If enabled, it is required to disable nm-cloud-setup and reboot the node:
 ```bash
 systemctl disable nm-cloud-setup.service nm-cloud-setup.timer
 reboot
+```
+
+### Ubuntu
+
+It is recommended to turn off ufw (uncomplicated firewall):
+```bash
+ufw disable
+```
+
+However, if you want to keep ufw enabled, you need to, at minimum, apply the following rules:
+```bash
+ufw allow from 10.42.0.0/16 to any
+ufw allow from 10.43.0.0/16 to any
 ```
 
 ### Raspberry Pi
--- a/docs/advanced/advanced.md
+++ b/docs/advanced/advanced.md
@@ -317,6 +317,14 @@ It is recommended to turn off firewalld:
 systemctl disable firewalld --now
 ```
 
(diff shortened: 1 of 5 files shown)


docs/advanced/advanced.md
```diff
@@ -316,12 +317,34 @@ It is recommended to turn off firewalld:
 systemctl disable firewalld --now
 ```
 
+If you wish to keep firewalld enabled, the following rules are required:
```
brandond, line 320: maybe note that these are the default pod and service CIDRs, and would need to be adjusted if the cidrs are changed?

docs/advanced/advanced.md
```diff
@@ -316,12 +317,34 @@ It is recommended to turn off firewalld:
...
+```bash
+firewall-cmd --permanent --add-port=6443/tcp # flannel
```
brandond, line 322: this is the supervisor/apiserver port. flannel also needs to be opened, as do the etcd ports if that's in use. Maybe cross-link to https://docs.k3s.io/installation/requirements#networking ?

docs/advanced/advanced.md
```diff
@@ -316,12 +317,34 @@ It is recommended to turn off firewalld:
...
+### Ubuntu
+
+It is recommended to turn off ufw (uncomplicated firewall):
+```bash
+ufw disable
+```
+
+If you wish to keep ufw enabled, the following rules are required:
+```bash
+ufw allow 6443/tcp
```
brandond, line 343: same notes as above

brandond (changes requested):

brandond (approved):

docs/installation/configuration.md
```diff
@@ -117,7 +117,7 @@ Finally, the location of the config file can be changed either through the CLI a
 Available as of [v1.21.0+k3s1](https://github.com/k3s-io/k3s/releases/tag/v1.21.0%2Bk3s1)
 :::
 
-Multiple configuration files are supported. Configuration is load by default from `/etc/rancher/k3s/config.yaml` and `/etc/rancher/k3s/config.d/*.yaml` in alphabetical order. The last value for a given key will be used. Slices are replaced.
+Multiple configuration files are supported. Configuration loads by default from `/etc/rancher/k3s/config.yaml` and `/etc/rancher/k3s/config.yaml.d/*.yaml` in alphabetical order. The last value for a given key will be used. Slices are replaced.
```
brandond, line 120:
```suggestion
Multiple configuration files are supported. By default, configuration files are read from `/etc/rancher/k3s/config.yaml` and `/etc/rancher/k3s/config.yaml.d/*.yaml` in alphabetical order. The last value for a given key will be used. A `+` can be appended to the key to append the value to the existing string or slice, instead of replacing it.
```

brandond (changes requested):

brandond (approved):

rbrtbnfgl (approved):

Klaas-:
is there an upside to using:
firewall-cmd --permanent --zone=trusted --add-source=10.42.0.0/16 #pods
firewall-cmd --permanent --zone=trusted --add-source=10.43.0.0/16 #services

instead of adding cni0 to trusted zone?

refs: https://github.com/k3s-io/k3s/issues/3122

If you want to keep this dependent on the sources I would suggest this should contain a reference to 10.42.0.0 and 10.43.0.0 being cluster-cidr and service-cidr and they are configureable, so if you change those options you also need to change them here.

brandond:
We do say that.

Klaas-:
Yeah I am blind :) I still would like to know if there is an upside to using the source instead of the interface, that way you'd get rid of the (potential) dynamic sources

dereknola:
The Pod and Services CIDR typically do not change once a cluster has been created. What your suggesting isn't really any better or worse than adding the specific CIDRS.

Issue #97, opened 2023-02-22
IPv6-only instructions are missing


rnett:
There is a [section](https://docs.k3s.io/installation/network-options#single-stack-ipv6-installation) for "Single-stack IPv6 installation", however there's no instructions on how to actually do it - the section is empty except for warnings.

brandond:
Don't use nodes with IPv4 addresses, and don't assign an IPv4 cluster CIDR. There's not really much else to it.

PR #98, opened 2023-02-24
Clarify Networking Documentation

dereknola:
Addresses:
- https://github.com/k3s-io/k3s/issues/1049
- https://github.com/k3s-io/k3s/issues/1310
- https://github.com/k3s-io/k3s/issues/1308
- https://github.com/k3s-io/docs/issues/97

Other Changes:
- Clarifies which versions added dual-stack and ipv6 only support.
- Adds more information about how to get logs.

### DIFF
--- a/docs/networking/networking.md
+++ b/docs/networking/networking.md
@@ -25,7 +25,7 @@ The Traefik ingress controller will use ports 80 and 443 on the host (i.e. these
 
 The `traefik.yaml` file should not be edited manually, because k3s would overwrite it again once it is restarted. Instead you can customize Traefik by creating an additional `HelmChartConfig` manifest in `/var/lib/rancher/k3s/server/manifests`. For more details and an example see [Customizing Packaged Components with HelmChartConfig](../helm/helm.md#customizing-packaged-components-with-helmchartconfig). For more information on the possible configuration values, refer to the official [Traefik Helm Configuration Parameters.](https://github.com/traefik/traefik-helm-chart/tree/master/traefik).
 
-To disable it, start each server with the `--disable traefik` flag.
+To disable it, start each server with the `--disable=traefik` flag.
 
 If Traefik is not disabled K3s versions 1.20 and earlier will install Traefik v1, while K3s versions 1.21 and later will install Traefik v2 if v1 is not already present.
 
--- a/docs/networking/networking.md
+++ b/docs/networking/networking.md
@@ -85,13 +85,16 @@ This is necessary if you wish to run a different LB, such as MetalLB.
 
 ## External Cloud Controller Manager
 
-K3s does not ship with any "in-tree" cloud providers. Instead, K3s ships with a Cloud Controller Manager (CCM) stub that does the following:
+In order to reduce binary size, K3s removes all "in-tree" (built-in) cloud providers. Instead, K3s provides an embedded Cloud Controller Manager (CCM) stub that does the following:
 - Sets node InternalIP and ExternalIP address fields based on the `--node-ip` and `--node-external-ip` flags.
 - Hosts the ServiceLB LoadBalancer controller.
 - Clears the `node.cloudprovider.kubernetes.io/uninitialized` taint that is present when the cloud-provider is set to `external` 
 
-In order to use an external CCM, you must start K3s with the `--disable-cloud-controller` flag. Additionally, if you disable the built-in CCM and do not provide an external CCM, nodes will remain tainted and unschedulable.
+Before deploying an external CCM, you must start all K3s servers with the `--disable-cloud-controller` flag to disable to embedded CCM. 
 
+:::note
+If you disable the built-in CCM and do not deploy and properly configure an external substitute, nodes will remain tainted and unschedulable.
+:::
 
 ## Nodes Without a Hostname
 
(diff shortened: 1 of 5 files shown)


docs/networking/networking.md
```diff
@@ -76,6 +83,14 @@ To disable the embedded LB, configure all servers in the cluster with the `--dis
 
 This is necessary if you wish to run a different LB, such as MetalLB.
 
+## External Cloud Controller Manager
+
+K3s ships with a stub Cloud Controller Manager (CCM) that does not actually implement any functionality. In order to use an external CCM, you must start K3s with the `--disable-cloud-controller` flag.
```
brandond, line 88: This isn't accurate. While it does not actually interface with any infrastructure providers, our stub CCM handles several critical operations:
* Setting node InternalIP and ExternalIP address fields based on the --node-ip and --node-external-ip flags.
* Hosting the ServiceLB LoadBalancer controller
* Clearing the `node.cloudprovider.kubernetes.io/uninitialized` taint that is present when the cloud-provider is set to `external` (which is the only valid value, since we disable all the built-in cloud providers).

If users disable the stub CCM, they must deploy an alternative CCM, or their nodes will remain tainted and unschedulable.

docs/networking/networking.md
```diff
@@ -76,6 +83,14 @@ To disable the embedded LB, configure all servers in the cluster with the `--dis
 
 This is necessary if you wish to run a different LB, such as MetalLB.
 
+## External Cloud Controller Manager
+
+K3s ships with a stub Cloud Controller Manager (CCM) that does not actually implement any functionality. In order to use an external CCM, you must start K3s with the `--disable-cloud-controller` flag. 
+
+:::note K3s v1.22 and older
+In order to deploy an external CCM with K3s v1.22 and older, you must also start K3s with the `--kubelet-arg="cloud-provider=external"` flag.
```
brandond, line 91: I don't recall this being necessary, did we not previously set that flag when the stub CCM is disabled?

brandond (changes requested):

docs/networking/networking.md
dereknola: This seems to have been necessary way back in v1.17 and older, I mangled that with the fact that in v1.23+ the `cloud-provider` arg is no longer a thing (its always external for kubelet). Removed!

docs/networking/networking.md
dereknola: Thanks for that info, I have incorporated it into the section.

docs/networking/networking.md
```diff
@@ -25,7 +25,7 @@ The Traefik ingress controller will use ports 80 and 443 on the host (i.e. these
 
 The `traefik.yaml` file should not be edited manually, because k3s would overwrite it again once it is restarted. Instead you can customize Traefik by creating an additional `HelmChartConfig` manifest in `/var/lib/rancher/k3s/server/manifests`. For more details and an example see [Customizing Packaged Components with HelmChartConfig](../helm/helm.md#customizing-packaged-components-with-helmchartconfig). For more information on the possible configuration values, refer to the official [Traefik Helm Configuration Parameters.](https://github.com/traefik/traefik-helm-chart/tree/master/traefik).
 
-To disable it, start each server with the `--disable traefik` option.
+To disable it, start each server with the `--disable traefik` flag.
```
brandond, line 28: nit: for consistency with how we suggest disabling servicelb down below
```suggestion
To disable it, start each server with the `--disable=traefik` flag.
```

docs/networking/networking.md
```diff
@@ -79,17 +79,19 @@ To select a particular subset of nodes to host pods for a LoadBalancer, set matc
...
 
-To disable the embedded LB, configure all servers in the cluster with the `--disable=servicelb` option.
+To disable the embedded LB, configure all servers in the cluster with the `--disable=servicelb` flag.
 
 This is necessary if you wish to run a different LB, such as MetalLB.
 
 ## External Cloud Controller Manager
 
-K3s ships with a stub Cloud Controller Manager (CCM) that does not actually implement any functionality. In order to use an external CCM, you must start K3s with the `--disable-cloud-controller` flag. 
+K3s does not ship with any "in-tree" cloud providers. Instead, K3s ships with a Cloud Controller Manager (CCM) stub that does the following:
```
brandond, line 88:
```suggestion
In order to reduce binary size, K3s removes all "in-tree" (built-in) cloud providers. Instead, K3s provides an embedded Cloud Controller Manager (CCM) that does the following:
```

docs/networking/networking.md
```diff
@@ -79,17 +79,19 @@ To select a particular subset of nodes to host pods for a LoadBalancer, set matc
...
+- Sets node InternalIP and ExternalIP address fields based on the `--node-ip` and `--node-external-ip` flags.
+- Hosts the ServiceLB LoadBalancer controller.
+- Clears the `node.cloudprovider.kubernetes.io/uninitialized` taint that is present when the cloud-provider is set to `external` 
+
+In order to use an external CCM, you must start K3s with the `--disable-cloud-controller` flag. Additionally, if you disable the built-in CCM and do not provide an external CCM, nodes will remain tainted and unschedulable.
```
brandond, line 93:
```suggestion
Before deploying an external CCM, you must start all K3s servers with the `--disable-cloud-controller` flag to disable the embedded CCM. 
:::note 
If you disable the embedded CCM and do not deploy and properly configure a substitute, nodes will remain tainted and unschedulable.
:::
```

brandond (changes requested):
couple wording suggestions

docs/networking/networking.md
brandond: 👍🏻 / 👎🏻 ?

dereknola: I added this in https://github.com/k3s-io/docs/pull/98/commits/8327f649a9dbed67a7f7fd0931f83ef56479381a

brandond (approved):

PR #101, opened 2023-03-02
Add enhance docs on certs and tokens

brandond:
* Use control-plane label instead of master in SUC docs
* Enhance docs on disabling things and using the deploy controller
* Add new section for CLI tools
* Add docs on custom CA certs and bootstrap tokens
* Reorg etcd backup, certificate rotate, and other CLI tools docs into new dedicated pages

Waiting on merge of https://github.com/k3s-io/k3s/pull/7032 to update utility scripts; links will be broken until then.

**Preview available at http://dev01.khaus.io:3000/**

## Linked Issues
* https://github.com/k3s-io/k3s/issues/2342
* https://github.com/k3s-io/k3s/issues/2645
* https://github.com/k3s-io/k3s/issues/5147
* https://github.com/k3s-io/k3s/issues/7012


dereknola (approved):
Are there K3s issues that we need to link to this PR?

docs/advanced/advanced.md
```diff
@@ -372,47 +357,37 @@ There are several ways to run K3s in Docker:
...
-[k3d](https://github.com/k3d-io/k3d) is a utility designed to easily run K3s in Docker.
-
-It can be installed via the [brew](https://brew.sh/) utility on MacOS:
-
-```bash
-brew install k3d
-```
-
-</TabItem>
-<TabItem value="Docker Compose">
```
dereknola, line 384: Why are we removing the docker compose tab? We still have a valid Docker compose available in the repo.

docs/advanced/advanced.md
```diff
@@ -11,40 +11,25 @@ import TabItem from '@theme/TabItem';
...
 
-### Automatic rotation
-By default, certificates in K3s expire in 12 months.
+### Certificate Authority Certificates
 
-If the certificates are expired or have fewer than 90 days remaining before they expire, the certificates are rotated when K3s is restarted.
+K3s generates self-signed Certificate Authority (CA) Certificates during startup of the first server node. These CA certificates are valid for 10 years, and are not automatically renewed.
 
-### Manual rotation
+For information on using custom CA certificates, or renewing the self-signed CA certificates, see the [`k3s certificate rotate-ca`](../reference/cli/certificate.md#rotate-ca) CLI tool documentation.
```
dereknola, line 20:
```suggestion
For information on using custom CA certificates, or renewing the self-signed CA certificates, see the [`k3s certificate rotate-ca` documentation](../reference/cli/certificate.md#rotate-ca).
```

I'm not in love with embedding the link within a code-block, because it hides the blue link color. So expanding the link to include "documentation" helps the user.

docs/advanced/advanced.md
```diff
@@ -11,40 +11,25 @@ import TabItem from '@theme/TabItem';
...
-
-Individual or lists of certificates can be rotated by specifying the certificate name:
-
-```bash
-k3s certificate rotate --service <SERVICE>,<SERVICE>
-```
+K3s client and server certificates are valid for 365 days from their date of issuance. Any certificates that are expired, or within 90 days of expiring, are automatically renewed every time K3s starts.
 
-The following certificates can be rotated: `admin`, `api-server`, `controller-manager`, `scheduler`, `k3s-controller`, `k3s-server`, `cloud-controller`, `etcd`, `auth-proxy`, `kubelet`, `kube-proxy`.
+For information on manually rotating client and server certificates, see the [`k3s certificate rotate`](../reference/cli/certificate.md#rotate) CLI tool documentation.
```
dereknola, line 26:
```suggestion
For information on manually rotating client and server certificates, see the [`k3s certificate rotate` documentation](../reference/cli/certificate.md#rotate).
```

docs/installation/addons.md
```diff
@@ -0,0 +1,62 @@
+---
```
dereknola, line 1: Rename this file to components.md or packaged-components.md. Its confusing that the name of the file is not in the Title.

docs/reference/cli/cli.md
```diff
@@ -0,0 +1,20 @@
+---
+title: CLI Tools
```
dereknola, line 2: Please move the whole CLI TOOLS folder up a level. Hiding it under Reference is not helpful. I specifically removed double deep folders when I reworked the docs website structure.

dereknola (changes requested):
This is more a comment for the future: Moving of things under the CLI TOOLs folder introduces a a clean break between a "reference" and a "guide". Move certain information back out of this folder would be helpful. For example: having the "Secrets Encryption" page contain a guide about how to do secrets rotation, while the CLI Tools SE page contains the hard information on every flag.

docs/security/secrets-encryption.md
```diff
@@ -52,258 +50,11 @@ Example of the encryption config file:
...
-  Stages are: `start`, `prepare`, `rotate`, `reencrypt_request`, `reencrypt_active`, `reencrypt_finished`  
-- __Server Encryption Hashes__: Useful for HA clusters, this indicates whether all servers are on the same stage with their local files. This can be used to identify whether a restart of servers is required before proceeding to the next stage. In the HA example above, node-1 and node-2 have different hashes, indicating that they currently do not have the same encryption configuration. Restarting the servers will sync up their configuration.
-- __Key Table__: Summarizes information about the secrets encryption keys found on the node.  
-  * __Active__: The "*" indicates which, if any, of the keys are currently used for secrets encryption. An active key is used by Kubernetes to encrypt any new secrets.
-  * __Key Type__: All keys using this tool are `AES-CBC` type. See more info [here.](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#providers)
-  * __Name__: Name of the encryption key.  
\ No newline at end of file
+For more information, see [`k3s secrets-encrypt`](../reference/cli/secrets-encrypt.md).
```
dereknola, line 60:
```suggestion
For more information, see [`k3s secrets-encrypt` documentation](../reference/cli/secrets-encrypt.md).
```

docs/advanced/advanced.md
brandond: As far as I know it's not maintained, tested, or supported. Running it directly via docker, or via k3d, are both better options than the compose file. I guess we can keep it around for now but I don't like steering people towards something that we don't actually recommend.

docs/advanced/advanced.md
brandond: that's fair. I can make sure that any where I'm linking with monospace there are also normal words included in the link.

docs/reference/cli/cli.md
brandond: Hmm. If we're going to do that, then I want to move the server and agent docs under this same section, as they are technically subcommands the same as everything else that we've got in this section.

I also moved the HA stuff under "Cluster Datastore Options", maybe we should move that up to the top-level and put the backup/restore stuff in with it.

docs/advanced/advanced.md
dereknola: If we don't want to support it, that's fine, we can remove it from the docs, but I think we should potentially remove it from master as well.

docs/reference/cli/cli.md
dereknola: +1 on moving the Cluster Datastore up to the top and adding the backup sections to it. 

I can add to the README in my next docs PR around convention.

docs/advanced/advanced.md
brandond: I'll open a PR to k3s to drop the docker-compose file.

brandond:
@dereknola 
> For example: having the "Secrets Encryption" page contain a guide about how to do secrets rotation, while the CLI Tools SE page contains the hard information on every flag.

Right now most of the CLI tools are pretty single-purpose, so having the flags documented on the same page as their usage makes the most sense to me. Only the main server and agent commands are complex enough, and intertwined with other things, that it makes sense to have a dedicated page just for all their CLI flags.

dereknola (approved):

cwayne18 (approved):

vickyhella:
Hi @brandond @dereknola, I noticed this PR also changed the CN docs lot, primarily by completely overriding the previous CN docs.
 
I am currently updating the K3s docs and found that it becomes hard for me to diff what was actually changed, and I need to rework the docs from scratch. Normally I only need to work on sentences that are changed.

So I was wondering if it's OK to leave the i18n folder for me to maintain? I will update the docs regularly and if there are any urgent update needs, you guys can let me know and I will prioritize it ^^

In addition, the previous translation is now all gone in production as follows:
image

PR #109, opened 2023-03-21
Update Chinese translation

vickyhella:
- Update Chinese translation


brandond (approved):
Thanks for getting this updated so quickly!

PR #110, opened 2023-03-21
Change from h1 to h3 so an anchor is generated

btat:
An anchor isn't generated for h1s so the section is not easily linkable. Updated to h3 to match the header level of other sections/entries on the page.


brandond:
I'm about to merge https://github.com/k3s-io/docs/pull/109 which already includes this

btat:
> I'm about to merge another PR which already includes this:
> 
>     * [Update Chinese translation #109](https://github.com/k3s-io/docs/pull/109)

Awesome, thanks for point that out.

PR #113, opened 2023-03-22
Update certificate.md

aw-engineer:
Updates "example script usage" filenames to match expectations of example script https://github.com/k3s-io/k3s/blob/master/contrib/util/generate-custom-ca-certs.sh

### DIFF
--- a/docs/cli/certificate.md
+++ b/docs/cli/certificate.md
@@ -112,13 +112,13 @@ If the files do not exist, the script will create new root and intermediate CA c
 :::
 
 If you want to use only an existing root CA certificate, provide the following files:
-* `root.pem`
-* `root.key`
+* `root-ca.pem`
+* `root-ca.key`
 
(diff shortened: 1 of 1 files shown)


brandond (changes requested):
Thanks for the PR! I've stared at this so many times my eyes have crossed and I completely missed that.

There are two additional places the filenames also need to be changed:
* https://github.com/k3s-io/docs/blob/4ad79ae19566c545d92e8bdfb7fc8bfe9c6afe95/docs/cli/certificate.md?plain=1#L131
* https://github.com/k3s-io/docs/blob/4ad79ae19566c545d92e8bdfb7fc8bfe9c6afe95/docs/cli/certificate.md?plain=1#L165

aw-engineer:
I have made the additional changes.  On L165, I think there needs to be a 'tls' directory, which I have also added.

brandond (approved):

PR #116, opened 2023-04-10
Fixed Typo in Configuration.md

MxCORU:
In the second line of  bash block  behind "When registering an agent, the following commands all result in the same behavior" was "K3s_TOKEN" instead of "K3S_TOKEN"


brandond:
Thanks for the PR!

PR #120, opened 2023-04-27
Fix flag for disabling servicelb

blachniet:
There is no `--disable-servicelb` flag. Instead, ServiceLB is disabled with the `--disable servicelb` flag and argument.

See [Using the `--disable` flag][1], **K3s Server CLI Help** at the bottom of the [`server` CLI page][2], and the [source code for the `server` flags][3].

[1]: https://docs.k3s.io/installation/packaged-components#using-the---disable-flag
[2]: https://docs.k3s.io/cli/server
[3]: https://github.com/k3s-io/k3s/blob/master/pkg/cli/cmds/server.go#L437

### DIFF
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -22,7 +22,7 @@ The following options must be set to the same value on all servers in the cluste
 * `--disable-cloud-controller`
 * `--disable-helm-controller`
 * `--disable-network-policy`
-* `--disable-servicelb`
+* `--disable servicelb`
 * `--egress-selector-mode`
 * `--flannel-backend`
 * `--flannel-external-ip`
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -22,7 +22,7 @@ The following options must be set to the same value on all servers in the cluste
 * `--disable-cloud-controller`
 * `--disable-helm-controller`
 * `--disable-network-policy`
-* `--disable servicelb`
+* `--disable=servicelb` *note: other packaged components may be disabled on a per-server basis*
 * `--egress-selector-mode`
 * `--flannel-backend`
 * `--flannel-external-ip`


docs/cli/server.md
```diff
@@ -22,7 +22,7 @@ The following options must be set to the same value on all servers in the cluste
 * `--disable-cloud-controller`
 * `--disable-helm-controller`
 * `--disable-network-policy`
-* `--disable-servicelb`
+* `--disable servicelb`
```
brandond, line 25: nit: Just delete the flag from here; the `--disable` flag is covered elsewhere.
```suggestion
```

brandond (changes requested):
Thanks for the PR! One nit.

docs/cli/server.md
blachniet: All of the flags in this section ([Critical Configuration Values][1]) are covered elsewhere, but it appears they are listed here because they "must be set to the same value on all servers in the cluster." I'm inclined to leave it in this list if it truly must be the same on all nodes (I don't know whether or not that it true for `--disable servicelb`), like the other flags listed here.

[1]: https://docs.k3s.io/cli/server#critical-configuration-values

brandond: No, `--disable` does not need to be the same on all servers.
https://github.com/k3s-io/k3s/blob/bc5b42c27908ab430101eff0db0a0b22f870bd7a/pkg/daemons/config/types.go#L132C1-L150

dereknola:
@blachniet Is this PR still relevant?

blachniet:
Hi @dereknola!

Yes, I believe this PR is still relevant. I'm sorry for my slow response, it fell off my radar. I applied @brandond's suggestion.

As @brandond pointed out, the [tag on the struct field](https://github.com/k3s-io/k3s/blob/bc5b42c27908ab430101eff0db0a0b22f870bd7a/pkg/daemons/config/types.go#L132C1-L150) is also incorrect. Should we address that in a separate issue? I'm not sure how to handle that. I'm guessing we can't put `disable=servicelb` in that struct tag, but I'm not familiar with that code.

brandond (approved):

dereknola (approved):

PR #125, opened 2023-05-07
Clarify method to access the dashboard from outside the cluster

nimbusparis:


dereknola:
This is already clearly described [in lines 61-85](https://github.com/k3s-io/docs/blob/main/docs/installation/kube-dashboard.md?plain=1#L61-L85) in the existing page

brandond:
I agree with @dereknola, this doesn't seem to be any different from the existing content, other than clarifying that you can run it on any host where you've got a copy of the admin kubeconfig and the `kubectl` binary - which should not be necessary to call out.

PR #134, opened 2023-05-31
Add cilium information and fix canal link

manuelbuil:
Add important information about Cilium to avoid problems when nuking the k3s cluster

Fix canal link which was broken

### DIFF
--- a/docs/installation/network-options.md
+++ b/docs/installation/network-options.md
@@ -52,12 +52,12 @@ We recommend that users migrate to the new backend as soon as possible. The migr
 
 ## Custom CNI
 
-Start K3s with `--flannel-backend=none` and install your CNI of choice. Most CNI plugins come with their own network policy engine, so it is recommended to set `--disable-network-policy` as well to avoid conflicts. IP Forwarding should be enabled for Canal and Calico; please reference the steps below.
+Start K3s with `--flannel-backend=none` and install your CNI of choice. Most CNI plugins come with their own network policy engine, so it is recommended to set `--disable-network-policy` as well to avoid conflicts. Some information is important to take into consideration:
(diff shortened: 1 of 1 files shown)


docs/installation/network-options.md
```diff
@@ -97,6 +97,19 @@ cat /etc/cni/net.d/10-calico.conflist
 You should see that IP forwarding is set to true.
 
 
+</TabItem>
+<TabItem value="Cilium" default>
+
+When nuking your k3s cluster, before running `k3s-killall.sh` or `k3s-uninstall.sh`, manually remove cilium_host, cilium_net and cilium_vxlan interfaces:
```
brandond, line 103: Put the consequences of failure up here, before the example command.
```suggestion
When before running `k3s-killall.sh` or `k3s-uninstall.sh`, you must manually remove the`cilium_host`, `cilium_net` and `cilium_vxlan` interfaces. If you fail to do this, you may loose network connectivity to the host when K3s is stopped.
```

docs/installation/network-options.md
```diff
@@ -97,6 +97,19 @@ cat /etc/cni/net.d/10-calico.conflist
...
+
+```bash
+ip link delete cilium_host
+ip link delete cilium_net
+ip link delete cilium_vxlan
+```
+
+Otherwise, your host network will hang and you will losa connectivity
+
```
brandond, line 112:
```suggestion
```

brandond (changes requested):

docs/installation/network-options.md
manuelbuil: thanks

brandond (approved):

PR #135, opened 2023-06-01
Rework section on Fixed Address Registration 

dereknola:
- Attempt to address issues talked about offline that resulted from https://github.com/k3s-io/docs/pull/95
- Bump docusaurus dependencies

### DIFF
--- a/docs/datastore/ha-embedded.md
+++ b/docs/datastore/ha-embedded.md
@@ -7,8 +7,15 @@ weight: 40
 Embedded etcd (HA) may have performance issues on slower disks such as Raspberry Pis running with SD cards.
 :::
 
-## New cluster
-To run K3s in this mode, you must have an odd number of server nodes. We recommend starting with three nodes.
+:::info
+HA embedded etcd cluster must be comprised of an odd number of server nodes for etcd to maintain quorum. An even number of server nodes will result in a non-functional, "split-brain" cluster. 
+:::
+
+An HA K3s cluster with embedded etcd is composed of:
+
+* Three or more **server nodes** that will serve the Kubernetes API and run other control plane services, as well as host the embedded etcd datastore.
+* Optional: Zero or more **agent nodes** that are designated to run your apps and services
+* Optional: A **fixed registration address** for agent nodes to register with the cluster
 
 To get started, first launch a server node with the `cluster-init` flag to enable clustering and a token that will be used as a shared secret to join additional servers to the cluster.
 ```bash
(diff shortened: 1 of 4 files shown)


docs/datastore/ha-embedded.md
```diff
@@ -7,8 +7,15 @@ weight: 40
 Embedded etcd (HA) may have performance issues on slower disks such as Raspberry Pis running with SD cards.
 :::
 
-## New cluster
-To run K3s in this mode, you must have an odd number of server nodes. We recommend starting with three nodes.
+:::info
+HA embedded etcd cluster must be comprised of an odd number of server nodes for etcd to maintain quorum. An even number of server nodes will result in a non-functional, "split-brain" cluster.
```
brandond, line 11: It wont split-brain if it loses quorum, it just won't work. We might quote from the etcd FAQ:

```suggestion
HA embedded etcd cluster must be comprised of an odd number of server nodes for etcd to maintain quorum. For a cluster with n servers, quorum is (n/2)+1. For any odd-sized cluster, adding one node will always increase the number of nodes necessary for quorum. Although adding a node to an odd-sized cluster appears better since there are more machines, the fault tolerance is worse since exactly the same number of nodes may fail without losing quorum but there are more nodes that can fail.
```

docs/datastore/ha.md
```diff
@@ -85,12 +73,24 @@ There are a few config flags that must be the same in all server nodes:
...
+
+* A layer-4 (TCP) load balancer
+* Round-robin DNS
+* Virtual or elastic IP addresses
+
+This endpoint can also be used for accessing the Kubernetes API. So you can, for example, modify your [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) file to point to it instead of a specific node. 
+
+To avoid certificate errors in such a configuration, you should install the server with the `--tls-san YOUR_IP_OR_HOSTNAME_HERE` option. This option adds an additional hostname or IP as a Subject Alternative Name in the TLS cert, and it can be specified multiple times if you would like to access via both the IP and the hostname.
```
brandond, line 96:
```suggestion
To avoid certificate errors in such a configuration, you should configure the server with the `--tls-san YOUR_IP_OR_HOSTNAME_HERE` option. This option adds an additional hostname or IP as a Subject Alternative Name in the TLS cert, and it can be specified multiple times if you would like to access via both the IP and the hostname.
```

brandond (changes requested):
nits

brandond (approved):

briandowns (approved):

PR #139, opened 2023-06-02
Update Resource Profiling to v1.26

dereknola:
- Testing K3s resources requirements with update K3s version and new AWS instances (C6 is now cheaper/better than C5). Addresses https://github.com/k3s-io/k3s/issues/3976
- Updated ARM 4K page requirement to indicate new versions don't need it. Addresses: https://github.com/k3s-io/docs/issues/137
- Add note about SSDs on Raspberry PIs

### DIFF
--- a/docs/reference/resource-profiling.md
+++ b/docs/reference/resource-profiling.md
@@ -39,9 +39,9 @@ The resource tests were intended to address the following problem statements:
 
 The tested components are:
 
-* K3s 1.19.2 with all packaged components enabled
+* K3s v1.26.5 with all packaged components enabled
 * Prometheus + Grafana monitoring stack
-* Kubernetes Example PHP Guestbook app
+* [Kubernetes Example Nginx Deployment](https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/)
 
 These are baseline figures for a stable system using only K3s packaged components (Traefik Ingress, Klipper lb, local-path storage) running a standard monitoring stack (Prometheus and Grafana) and the Guestbook example app.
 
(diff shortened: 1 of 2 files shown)


brandond (approved):

docs/reference/resource-profiling.md
```diff
@@ -39,83 +39,95 @@ The resource tests were intended to address the following problem statements:
...
+|               | Pi4B |  1413 M |
 
 ### K3s Agent
 
-The CPU requirements are:
+The requirements are:
+
+| System | CPU Core Usage | RAM |
+|--------|----------------|-----|
+| c6id.xlarge | 3% of a core | 275 M |
```
brandond, line 126: Can we keep this as a CPU model/speed? I have no idea what a c6id instance uses for the processor.

brandond (changes requested):

docs/reference/resource-profiling.md
```diff
@@ -9,12 +9,12 @@ The results are summarized as follows:
...
 | Components |  Processor | Min CPU | Min RAM with Kine/SQLite | Min RAM with Embedded etcd |
 |------------|-----|----------|-------------------------|---------------------------|
-| K3s server with a workload  | Intel(R) Xeon(R) Platinum 8124M CPU, 3.00 GHz | 10% of a core | 768 M | 896 M |
-| K3s cluster with a single agent  | Intel(R) Xeon(R) Platinum 8124M CPU, 3.00 GHz | 10% of a core | 512 M | 768 M |
-| K3s agent | Intel(R) Xeon(R) Platinum 8124M CPU, 3.00 GHz | 5% of a core | 256 M | 256 M |
-| K3s server with a workload  | Pi4B BCM2711, 1.50 GHz | 20% of a core | 768 M | 896 M |
-| K3s cluster with a single agent | Pi4B BCM2711, 1.50 GHz | 20% of a core | 512 M | 768 M |
-| K3s agent  | Pi4B BCM2711, 1.50 GHz | 10% of a core | 256 M | 256 M |
+| K3s server with a workload  | Intel 8375C CPU, 2.90 GHz | 6% of a core | 1602 M | 1604 M |
+| K3s cluster with a single agent  | Intel 8375C CPU, 2.90 GHz | 5% of a core | 1428 M | 1450 M |
```
brandond, line 13: The memory utilization for kine was within 2M of etcd for both test cases? That is highly unexpected and seems unusual; etcd usually uses a noticeable chunk more memory.

dereknola: Retested 1 Server workload:
kine before workload: 1419M, after: 1596M
etcd before workload: 1425M, after: 1606M

On the x86 arch there is not that much difference between the DBs. I note that this is always "at rest" after installation/addition of the workload. There is a consistent garbage collection pass that tends to drop the memory usage after about 1 min. Its possible that kine/sqlite has less usage during spikes or in more active clusters.

brandond (approved):

ShylajaDevadiga (approved):

PR #142, opened 2023-06-13
Add tailscale docs

manuelbuil:
Adding tailscale docs. Tailscale integration will be out in the June releases

### DIFF
--- a/docs/installation/network-options.md
+++ b/docs/installation/network-options.md
@@ -182,10 +182,21 @@ Single-stack IPv6 clusters (clusters without IPv4) are supported on K3s using th
 --cluster-cidr=2001:cafe:42:0::/56 --service-cidr=2001:cafe:42:1::/112
 ```
 
-
 ## Distributed hybrid or multicloud cluster
 
-A K3s cluster can still be deployed on nodes which use different private networks and are not directly connected (e.g. nodes in different public clouds). To achieve this, K3s sets a mesh of tunnels that become a VPN mesh. These nodes must have have an assigned IP through which they can be reached (e.g. a public IP). The server traffic will use a websocket tunnel and the data-plane traffic will use a wireguard tunnel.
+A K3s cluster can still be deployed on nodes which do not share a common private network and are not directly connected (e.g. nodes in different public clouds). There are two options to achieve this: the embedded k3s multicloud solution and the integration with the `tailscale` VPN provider.
+
+:::caution Warning
+The latency between nodes will increase as external connectivity requires more hops. This will reduce the network performance and could also impact the health of the cluster if latency is too high.
+:::
+
+:::caution Warning
+Embedded etcd is not supported in this type of deployment. If using embedded etcd;, all server nodes must be reachable to each other via their private IPs. Agents may be distributed over multiple networks, but all servers should be in the same location.
+:::
+
+### Embedded k3s multicloud solution
+
+K3s uses wireguard to establish a VPN mesh for cluster traffic. Nodes must each have a unique IP through which they can be reached (usually a public IP). K3s supervisor traffic will use a websocket tunnel, and cluster (CNI) traffic will use a wireguard tunnel.
 
 To enable this type of deployment, you must add the following parameters on servers:
 ```bash
@@ -201,7 +212,7 @@ where `SERVER_EXTERNAL_IP` is the IP through which we can reach the server node 
 Both `SERVER_EXTERNAL_IP` and `AGENT_EXTERNAL_IP` must have connectivity between them and are normally public IPs.
 
 :::info Dynamic IPs
-If nodes are assigned dynamic IPs and the IP changes (e.g. in AWS), you must modify the `--node-external-ip` parameter to reflect the new IP. If running k3s as a service, you must modify `/etc/systemd/system/k3s.service` then run:
+If nodes are assigned dynamic IPs and the IP changes (e.g. in AWS), you must modify the `--node-external-ip` parameter to reflect the new IP. If running K3s as a service, you must modify `/etc/systemd/system/k3s.service` then run:
 
 ```bash
 systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


docs/installation/network-options.md
```diff
@@ -182,10 +182,21 @@ Single-stack IPv6 clusters (clusters without IPv4) are supported on K3s using th
 --cluster-cidr=2001:cafe:42:0::/56 --service-cidr=2001:cafe:42:1::/112
 ```
 
-
 ## Distributed hybrid or multicloud cluster
 
-A K3s cluster can still be deployed on nodes which use different private networks and are not directly connected (e.g. nodes in different public clouds). To achieve this, K3s sets a mesh of tunnels that become a VPN mesh. These nodes must have have an assigned IP through which they can be reached (e.g. a public IP). The server traffic will use a websocket tunnel and the data-plane traffic will use a wireguard tunnel.
+A K3s cluster can still be deployed on nodes which use different private networks and are not directly connected (e.g. nodes in different public clouds). There are two options to achieve this: wireguard-native and tailscale
```
brandond, line 187:
```suggestion
A K3s cluster can be deployed on nodes which do not share a common private network and are not directly connected (e.g. nodes in different public clouds). There are two options to achieve this: the `wireguard-native` flannel backend, and the `tailscale` VPN provider.
```

docs/installation/network-options.md
```diff
@@ -182,10 +182,21 @@ Single-stack IPv6 clusters (clusters without IPv4) are supported on K3s using th
...
+
+:::caution Warning
+The latency between nodes will increase as external connectivity requires more hops. This will reduce the network performance and could also impact the health of the cluster if latency is too high.
+:::
+
+:::caution Warning
+Embedded etcd is not supported in this type of deployment. If using embedded etcd; all server nodes must be reachable to each other via their private IPs.
```
brandond, line 194:
```suggestion
If using embedded etcd; all server nodes must be reachable to each other via their private IPs. Agents may be distributed over multiple networks, but all servers should be in the same location.
```

docs/installation/network-options.md
```diff
@@ -182,10 +182,21 @@ Single-stack IPv6 clusters (clusters without IPv4) are supported on K3s using th
...
+:::
+
+### Wireguard-native
+
+K3s can set a mesh of tunnels that become a VPN mesh. Nodes must have have an assigned IP through which they can be reached (e.g. a public IP). The server traffic will use a websocket tunnel and the data-plane traffic will use a wireguard tunnel.
```
brandond, line 199:
```suggestion
K3s use wireguard to establish a VPN mesh for cluster traffic. Nodes must each have a unique IP through which they can be reached (usually a public IP). K3s supervisor traffic will use a websocket tunnel, and cluster (CNI) traffic will use a wireguard tunnel.
```

docs/installation/network-options.md
```diff
@@ -209,11 +220,31 @@ systemctl restart k3s
...
-The latency between nodes will increase as external connectivity requires more hops. This will reduce the network performance and could also impact the health of the cluster if latency is too high.
-:::
+### Tailscale
 
-:::caution Warning
-Embedded etcd will not use external IPs for communication. If using embedded etcd; all server nodes must be reachable to each other via their private IPs.
-:::
+Available in v1.27.3, v1.26.6, v1.25.11 and newer.
+
+K3s can integrate with [tailscale](https://tailscale.com/) so that nodes use the tailscale VPN network to communicate with each other.
```
brandond, line 227:
```suggestion
K3s can integrate with [Tailscale](https://tailscale.com/) so that nodes use the Tailscale VPN service to build a mesh between nodes.
```

docs/installation/network-options.md
```diff
@@ -209,11 +220,31 @@ systemctl restart k3s
...
+
+There are four steps to be done with tailscale before deploying k3s:
+1 - Create an account in tailscale
```
brandond, line 230:
```suggestion
There are four steps to be done with Tailscale before deploying k3s:
1 - Log in to your Tailscale account
```

docs/installation/network-options.md
```diff
@@ -209,11 +220,31 @@ systemctl restart k3s
...
+2 - In Settings > Keys, generate an auth key (<auth-key>), which may be reusable for all nodes in your cluster
```
brandond, line 231: Does this need to be unique per cluster? Can multiple clusters use the same key? Are there any other considerations for having multiple clusters on a single account?

docs/installation/network-options.md
```diff
@@ -209,11 +220,31 @@ systemctl restart k3s
...
+```yaml
+"autoApprovers": {
+		"routes": {
+			"10.42.0.0/16":        ["your_account@xyz.com"],
+			"2001:cafe:42:0::/56": ["your_account@xyz.com"],
+		},
+	},
+```
+
+4 - Install tailscale in your nodes:
```
brandond, line 242:
```suggestion
4 - Install Tailscale in your nodes:
```

docs/installation/network-options.md
```diff
@@ -209,11 +220,31 @@ systemctl restart k3s
...
+```bash
+curl -fsSL https://tailscale.com/install.sh | sh
+```
 
+To deploy k3s with tailscale integration enabled, you must add the following parameter on each of your nodes:
```
brandond, line 247:
```suggestion
To deploy k3s with Tailscale integration enabled, you must add the following parameter on each of your nodes:
```

brandond (changes requested):
a couple wording suggestions, plus we should be consistent about capitalizing Tailscale when referring to the product name.

docs/installation/network-options.md
manuelbuil: ok, thanks!

docs/installation/network-options.md
manuelbuil: This is not accurate though, because one option is not `wireguard-native`. That is what pod traffic is using, however, this is not what control-planes are using. They use public IP + https to communicate. So, summarizing the solution by saying `wireguard-native` is misleading. I think we should find a better naming, something like: 
```
There are two options to achieve this: the k3s internal multicloud solution, and the integration with tailscale VPN provider.
```
And then describe both

docs/installation/network-options.md
manuelbuil: ok

docs/installation/network-options.md
manuelbuil: good, thanks!

docs/installation/network-options.md
manuelbuil: English question: `Log in to` or `log into`

docs/installation/network-options.md
manuelbuil: :+1:

docs/installation/network-options.md
brandond: Maybe "Sign in"? I have seen it as "login to", " log in to", and "log into", but I think only the three word version is grammatically correct.

brandond (approved):

dereknola (approved):

brandond:
Ugh the ordered list didn't come out right here:
https://docs.k3s.io/installation/network-options#integration-with-the-tailscale-vpn-provider-experimental

manuelbuil:
> Ugh the ordered list didn't come out right here: https://docs.k3s.io/installation/network-options#integration-with-the-tailscale-vpn-provider-experimental
> 
> These should have been a proper markdown list:
> 
> ```gfm
> 1. one
> 2. two
> 3. three
> ```
> 
> There are also a couple odd bits of punctuation that I missed.

You are right! My day was too busy yesterday to look into this. I have time today, let me prepare a second PR to improve it

PR #143, opened 2023-06-15
Improve tailscale docs

manuelbuil:
This is how it looks now:

image


brandond (changes requested):
There's a punctuation issue on line 194 as well: `embedded etcd;, all`

brandond (approved):

PR #148, opened 2023-06-29
Change <NAMESPACE> to kube-system

AdoPi:
Hello!
I'm not sure why in the hardening docs there is a namespace named `<NAMESPACE>`.
After reading the doc, I think `kube-system` should be used as an example instead of  `<NAMESPACE>`.

### DIFF
--- a/docs/security/hardening-guide.md
+++ b/docs/security/hardening-guide.md
@@ -395,7 +395,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: default-network-dns-policy
-  namespace: <NAMESPACE>
+  namespace: kube-system
 spec:
   ingress:
   - ports:


docs/security/hardening-guide.md
```diff
@@ -395,7 +395,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: default-network-dns-policy
-  namespace: <NAMESPACE>
```
brandond, line 398: No, don't do this - it's a placeholder. You'll need a policy like this in any namespace that wants to use DNS; you're supposed to replace `<NAMESPACE>` with the actual namespace of the pods that you want to be able to use DNS.

Traffic within the kube-system namespace is handled by the policy above this one, and doesn't need a specific policy to reach the DNS service.

If you wanted to enhance the comment above this example policy to explain that it's a placeholder, that would be great - but don't remove the placeholder value.

brandond (changes requested):

docs/security/hardening-guide.md
AdoPi: Ok I see thanks, I guess I did something wrong because if I don't set namespace: kube-system my pods can't access to DNS.

PR #150, opened 2023-07-01
Update manual.md

GeoffWilliams:
added `-s` to curl/bash commands to fix command error on debian 12, eg:

```
geoff@cloud:~$ curl -sfL https://get.k3s.io | sh - --disable traefik --disable servicelb
sh: 0: cannot open --disable: No such file
```


brandond (approved):
thanks!

PR #152, opened 2023-07-12
Add guide on external load balancers

dereknola:
## Changes
- Add dependabot to manage yarn dependecies 
- Remove all old pages, added automatic redirects
- Add guide on setting up an external loadbalancer with HAProxy or Nginx Community

Addresses: https://github.com/k3s-io/k3s/issues/5126

Currently hosted at: https://dereknola.github.io/docs-k3s/datastore/cluster-loadbalancer

### DIFF
--- a/docs/datastore/cluster-loadbalancer.md
+++ b/docs/datastore/cluster-loadbalancer.md
@@ -1,0 +1,137 @@
+---
+title: Cluster Load Balancer
+weight: 30
+---
+
+
+This section describes how to install an external load balancer for use with a High Availability (HA) K3s cluster. Two examples are provided: Nginx and HAProxy.
+
+:::tip
+An external loadbalancer should not be confused with the embedded ServiceLB, which is an internal load balancer that allows pods to maintain static IPs inside the cluster itself. For more details, see [Service Load Balancer](../networking/networking.md#service-load-balancer).
+:::
+
+## Prerequisites
+
+For both examples, assume that a [HA K3s cluster with embedded etcd](../datastore/ha-embedded.md) has been installed on 3 nodes.
+
+Each k3s server is configured with:
+```yaml
+# /etc/rancher/k3s/config.yaml
+token: lb-cluster-gd
+tls-san: 10.10.10.99
+```
+
+The nodes have hostnames and IPs of: 
+* server-1: `10.10.10.50`
+* server-2: `10.10.10.51`
+* server-3: `10.10.10.52`
+
+
+An additional node with docker installed is also configured, with hostname and IP of:
+* lb-1: `10.10.10.99`
+
+Three additional nodes exist with hostnames and IPs of:
+* agent-1: `10.10.10.101`
+* agent-2: `10.10.10.102`
+* agent-3: `10.10.10.103`
+
+## Nginx Load Balancer
+
+[Nginx Open Source](http://nginx.org/) provides a TCP load balancer. See [Using nginx as HTTP load balancer](https://nginx.org/en/docs/http/load_balancing.html) for more info.
+
+1) Create a `nginx.conf` file on lb-1 with the following contents:
+
+```
+events {}
+
+stream {
+  upstream k3s_servers {
+    server 10.10.10.50:6443;
+    server 10.10.10.51:6443;
+    server 10.10.10.52:6443;
+  }
+
+  server {
+    listen 6443;
+    proxy_pass k3s_servers;
+  }
+}
+```
+
+2) Run the following command on lb-1 to start the Nginx load balancer:
+
+```bash
+docker run -d --restart unless-stopped \
+    -v ${PWD}/nginx.conf:/etc/nginx/nginx.conf \
+    -p 6443:6443 \
+    nginx:stable
+```
+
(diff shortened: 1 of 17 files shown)


brandond (changes requested):
I'm a little confused why we would want to instruct users on how to add an external single-node load-balancer in front of k3s. This is just as much of a single point of failure as simply pointing clients at one of the servers.

If we're going do do this, I feel like we should prominently warn them that this is not any more HA than just using a single K3s server as the designated endpoint for external clients.

dereknola:
The idea is to flesh out sections such as https://docs.k3s.io/architecture#high-availability-k3s and https://docs.k3s.io/datastore/ha#5-optional-configure-a-fixed-registration-address with some example. In most of our architecture pictures we have a loadbalancer of some kind placed in front of the agents.

brandond:
I think the expectation is that it's a cloud provider's load-balancer service, where the HA is managed by a 3rd party. That is the most common case that I see our users doing. IMO there's not really much value in putting another single node / single point of failure in front of a HA K3s cluster.

brandond:
Yeah, that'd be great. We could probably still include a single-node example, I would just want to preface it with a lot of warnings about single points of failure. We could probably also be more explicit about encouraging use of an external HA load-balancer service whenever possible.

dereknola:
@brandond Check out https://dereknola.github.io/docs-k3s/datastore/cluster-loadbalancer with the page rework.

docs/datastore/cluster-loadbalancer.md
```diff
@@ -0,0 +1,192 @@
+---
+title: Cluster Load Balancer
+weight: 30
+---
+
+
+This section describes how to install an external load balancer for use with a High Availability (HA) K3s cluster. Two examples are provided: Nginx and HAProxy.
+
+:::tip
+An external loadbalancer should not be confused with the embedded ServiceLB, which is an internal load balancer that allows pods to maintain static IPs inside the cluster itself. For more details, see [Service Load Balancer](../networking/networking.md#service-load-balancer).
```
brandond, line 10:
```suggestion
An external load-balancers should not be confused with the embedded ServiceLB, which is an embedded controller that allows for use of Kubernetes LoadBalancer Services without deploying a third-party load-balancer controller. For more details, see [Service Load Balancer](../networking/networking.md#service-load-balancer). 

External load-balancers can be used to provide a fixed registration address for registering nodes, or for external access to the Kubernetes API Server. For exposing LoadBalancer Services, external load-balancers can be used alongside or instead of ServiceLB, but in most cases, replacement load-balancer controllers such as MetalLB or Kube-VIP are a better choice.

File: docs/datastore/cluster-loadbalancer.md [hunk 2]
```diff
@@ -0,0 +1,192 @@
+---
+title: Cluster Load Balancer
+weight: 30
+---
+
+
+This section describes how to install an external load balancer for use with a High Availability (HA) K3s cluster. Two examples are provided: Nginx and HAProxy.
```
brandond, line 7:
```suggestion
This section describes how to install an external load balancer in front of a High Availability (HA) K3s cluster's server nodes. Two examples are provided: Nginx and HAProxy.
```

docs/datastore/cluster-loadbalancer.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+</TabItem>
+
+<TabItem value="Nginx">
+
+## Nginx Load Balancer
+
+:::warning
+Nginx does not natively support a High Avaliability (HA) configuration. If setting up an HA cluster, having a single loadbalancer node will reintroduce a single point of failure.
```
brandond, line 132:
```suggestion
Nginx does not natively support a High Availability (HA) configuration. If setting up an HA cluster, having a single load balancer in front of K3s will reintroduce a single point of failure.
```

brandond (changes requested):

docs/datastore/cluster-loadbalancer.md
```diff
@@ -7,7 +7,9 @@ weight: 30
 This section describes how to install an external load balancer for use with a High Availability (HA) K3s cluster. Two examples are provided: Nginx and HAProxy.
 
 :::tip
-An external loadbalancer should not be confused with the embedded ServiceLB, which is an internal load balancer that allows pods to maintain static IPs inside the cluster itself. For more details, see [Service Load Balancer](../networking/networking.md#service-load-balancer).
+An external load-balancers should not be confused with the embedded ServiceLB, which is an embedded controller that allows for use of Kubernetes LoadBalancer Services without deploying a third-party load-balancer controller. For more details, see [Service Load Balancer](../networking/networking.md#service-load-balancer).
```
brandond, line 10: Minor grammar nit, lgtm otherwise

```suggestion
External load-balancers should not be confused with the embedded ServiceLB, which is an embedded controller that allows for use of Kubernetes LoadBalancer Services without deploying a third-party load-balancer controller. For more details, see [Service Load Balancer](../networking/networking.md#service-load-balancer). 
```

brandond (changes requested):

brandond (approved):

briandowns (approved):

PR #154, opened 2023-07-17
Give example of append key to stringslice

dereknola:
We mention using the `+` in a config but don't give a good example if it.

### DIFF
--- a/docs/installation/configuration.md
+++ b/docs/installation/configuration.md
@@ -122,7 +122,9 @@ Finally, the location of the config file can be changed either through the CLI a
 Available as of [v1.21.0+k3s1](https://github.com/k3s-io/k3s/releases/tag/v1.21.0%2Bk3s1)
 :::
 
-Multiple configuration files are supported. By default, configuration files are read from `/etc/rancher/k3s/config.yaml` and `/etc/rancher/k3s/config.yaml.d/*.yaml` in alphabetical order. The last value for a given key will be used. A `+` can be appended to the key to append the value to the existing string or slice, instead of replacing it.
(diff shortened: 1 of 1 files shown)


brandond (commented):
This is great, we did need an example.

I'm not sure if there's a good place to call this out, but when using the `key+` syntax, the order of the files does matter - for example:

```yaml
# test1.yaml
node-taint:
  - alice=bob:NoExecute
```
```yaml
# test2.yaml
node-taint+:
  - charlie=delta:NoSchedule
```
```yaml
# test3.yaml
node-taint:
  - eve=frank:NoExecute
```

The value from the 3rd file will clobber the combined values from the first two, as the files are loaded in lexicographic order, and appending happens when the file is loaded - it does not forever taint that key as an accumulating key. Maybe it should, I don't know.

dereknola:
I think this is pretty clear from the documentation wording already, we say:

>  By default, configuration files are read from /etc/rancher/k3s/config.yaml and /etc/rancher/k3s/config.yaml.d/*.yaml in alphabetical order. The last value for a given key will be used. A + can be appended to the key to append the value to the existing string or slice, instead of replacing it.

Maybe I could reword it to:
```
By default, configuration files are read from /etc/rancher/k3s/config.yaml and /etc/rancher/k3s/config.yaml.d/*.yaml in alphabetical order. 

By default, the last value found for a given key will be used. A `+` can be appended to the key to append the value to the existing string or slice, instead of replacing it.
All subsequent entries of this key will also require a `+` to prevent overwriting the entire value.
```

brandond (approved):

briandowns (approved):

PR #182, opened 2023-09-29
Added IPv6 on the advertise-address description

rbrtbnfgl:
Added IPv6 also on the `advertise-address` description because is a valid value in case IPv6 is the primary address for the service CIDR.

### DIFF
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -96,7 +96,7 @@ The following options must be set to the same value on all servers in the cluste
 | --------------------------- | ------------------------ | -------------------------------------------------------------------------------------------- |
 | `--bind-address` value      | 0.0.0.0                  | k3s bind address                                                                             |
 | `--https-listen-port` value | 6443                     | HTTPS listen port                                                                            |
-| `--advertise-address` value | node-external-ip/node-ip | IPv4 address that apiserver uses to advertise to members of the cluster                      |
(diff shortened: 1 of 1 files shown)


docs/cli/server.md
```diff
@@ -96,7 +96,7 @@ The following options must be set to the same value on all servers in the cluste
 | --------------------------- | ------------------------ | -------------------------------------------------------------------------------------------- |
 | `--bind-address` value      | 0.0.0.0                  | k3s bind address                                                                             |
 | `--https-listen-port` value | 6443                     | HTTPS listen port                                                                            |
-| `--advertise-address` value | node-external-ip/node-ip | IPv4 address that apiserver uses to advertise to members of the cluster                      |
+| `--advertise-address` value | node-external-ip/node-ip | IPv4/6 address that apiserver uses to advertise to members of the cluster (note that the primary service cluster IP range should be of the same family of the advertised address) |
```
brandond, line 99: We use `IPv4/IPv6` for other flag descriptions, and a nit on the wording.
I think the note will also look better on a separate line
```suggestion
| `--advertise-address` value | node-external-ip/node-ip | IPv4/IPv6 address that apiserver advertises for its service endpoint<br>Note that the primary `service-cidr` IP range must be of the same address family as the advertised address |
```

brandond (changes requested):

brandond (approved):

manuelbuil (approved):

PR #185, opened 2023-10-11
Update quick-start.md

NayukiLeung:
Add instructions for verifying installation and answers to questions.


dereknola (changes requested):
I don't think its necessary to put this info in the Quick Start page. That page is designed to get a user up and running and avoid complex detail of how K3s could/should work. A better place for this info would be the Cluster Access page.

NayukiLeung:
> I don't think its necessary to put this info in the Quick Start page. That page is designed to get a user up and running and avoid complex detail of how K3s could/should work. A better place for this info would be the Cluster Access page.

Thanks for your advice : )

dereknola:
Closed due to inactivity. Feel free to comment and reopen.

PR #186, opened 2023-10-11
Update server.md - make "K3s Server CLI Help" header a hyper link

jessebot:
Previously, "K3s Server CLI Help" was not a hyperlink, and I think that's because it was prefixed with `#`, which is used for the title of the entire page.

I've changed it to `##` so that it will be header 2, and should show up as a clickable link and on the right hand side table of contents.


brandond:
@dereknola we need to exclude the release notes from spellcheck, or retroactively fix them I guess...

brandond (approved):

dereknola (approved):

PR #188, opened 2023-10-16
Networking content update

manuelbuil:
This PR does the following:
* Removes the option `--flannel-backend=wireguard` (deprecated since 1.24 and removed in 1.26)
* Hides the section "Migrating from `wireguard` or `ipsec` to `wireguard-native`" from the side summary
* Updates warning about dualStack node-ip
* Adds known issue when doing dualStack with IPv6 first
* Removes Multicluster CIDR
* Fixed this error thrown by CI:
```
./docs/release-notes/v1.27.X.md:134: porfile ==> profile
Error: Process completed with exit code 6[5](https://github.com/k3s-io/docs/actions/runs/6532077574/job/17734575997?pr=188#step:5:6).
```

### DIFF
--- a/docs/installation/network-options.md
+++ b/docs/installation/network-options.md
@@ -27,7 +27,6 @@ This page describes K3s network configuration options, including configuration o
  `--flannel-backend=host-gw` | Use IP routes to pod subnets via node IPs. Requires direct layer 2 connectivity between all nodes in the cluster. |
  `--flannel-backend=wireguard-native` | Use WireGuard to encapsulate and encrypt network traffic. May require additional kernel modules. |
  `--flannel-backend=ipsec` | Use strongSwan IPSec via the `swanctl` binary to encrypt network traffic. (Deprecated; will be removed in v1.27.0) |
- `--flannel-backend=wireguard` | Use WireGuard via the `wg` binary to encrypt network traffic. May require additional kernel modules and configuration. (Deprecated; will be removed in v1.26.0) |
  `--flannel-backend=none` | Disable Flannel entirely. |
 
 :::info Version Gate
@@ -36,7 +35,7 @@ K3s no longer includes strongSwan `swanctl` and `charon` binaries starting with 
 
 :::
 
-### Migrating from `wireguard` or `ipsec` to `wireguard-native`
+#### Migrating from `wireguard` or `ipsec` to `wireguard-native`
 
 The legacy `wireguard` backend requires installation of the `wg` tool on the host. This backend will be removed in K3s v1.26, in favor of `wireguard-native` backend, which directly interfaces with the kernel.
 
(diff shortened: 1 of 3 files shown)


rbrtbnfgl:
does the update of yarn.lock file needed?

manuelbuil:
> does the update of yarn.lock file needed?

good catch! And the json is also not needed

rbrtbnfgl (approved):

brandond:
Do we need a Table Of Contents at the top, in addition to the auto-generated one on the side? Could we perhaps just clean up the section titles so that the auto-generated TOC is more usable?

image


A lot of the stuff on this page isn't really related to networking - coredns and traefik for example should probably be covered elsewhere, leaving this page for just actual netpol/flannel/dual-stack/vpn content.

manuelbuil:
> Do we need a Table Of Contents at the top, in addition to the auto-generated one on the side? Could we perhaps just clean up the section titles so that the auto-generated TOC is more usable?
> 
> image
> 
> A lot of the stuff on this page isn't really related to networking - coredns and traefik for example should probably be covered elsewhere, leaving this page for just actual netpol/flannel/servicelb/dual-stack/vpn content, which could be reorganized to be easier to navigate.

My eyes were ignoring the sidebar Table of Contents, that's why I added it :). But now that I see it, I think it is repetitive. Instead I am doing just a regular update of the content.

I will create an issue to reorg our networking content which is confusing. It is covered in two different places and I think that is confusing but it is also not obvious how to divide it

docs/installation/network-options.md
```diff
@@ -163,6 +162,10 @@ Note that you may configure any valid `cluster-cidr` and `service-cidr` values,
 
 If you are using a custom CNI plugin, i.e. a CNI plugin other than Flannel, the additional configuration may be required. Please consult your plugin's dual-stack documentation and verify if network policies can be enabled.
 
+:::caution Known Issue
+When defining cluster-cidr and service-cidr, if the first IPFamily is IPv6, node-ip should be specified if you want the main IP of the pods to be IPv6, otherwise kubelet will always pick IPv4.
```
brandond, line 166:
```suggestion
When defining cluster-cidr and service-cidr with IPv6 as the primary address family, the node-ip of all cluster members should be explicitly set to the node's desired IPv6 address if you want the primary IP of pods to be IPv6. By default, the kubelet will always use IPv4 as the primary address family.
```

docs/networking/networking.md
```diff
@@ -3,7 +3,7 @@ title: "Networking"
 weight: 35
 ---
 
-This page explains how CoreDNS, the Traefik Ingress controller, and Klipper service load balancer work within K3s.
+This page explains how CoreDNS, the Traefik Ingress controller, Network Policy Controller and Klipper service load balancer work within K3s.
```
brandond, line 6: Oxford comma. Also, I think we're not calling stuff Klipper so much any more.
```suggestion
This page explains how CoreDNS, Traefik Ingress controller, Network Policy controller, and ServiceLB load balancer controller work within K3s.
```

brandond (changes requested):

docs/installation/network-options.md
manuelbuil: thanks! Changed your suggestion slightly

docs/networking/networking.md
manuelbuil: thanks!

brandond (approved):

thomasferrandiz (approved):

rbrtbnfgl (approved):

PR #199, opened 2023-11-16
Move Advanced OS preop into requirements

dereknola:
Addresses: https://github.com/k3s-io/docs/issues/195

## Changes:
- Move Additonal OS prep in collapsed sections in the Requirements page
- Add blurb in quickstart linking to requirments page
- Move Iptables workaround fully into Known Issues page


brandond:
Should we do tabs like we do for the different database engines, instead of collapsable sections?

dereknola:
So originall in Docusaurus 2.0, it was not possible to have any kind of linking or anchoring for items inside tabs. Thats why I  was using the details collapse, because then we could still link direct to Raspberry-Pi in several places. However now that https://github.com/facebook/docusaurus/pull/8225 is merged, Docusaurus 2.3 allows using query-strings allows some level of direct linking, so I will try that out.

VestigeJ (approved):
LGTM

brandond (approved):
love it!

PR #204, opened 2023-11-21
Add section on node-taints

dereknola:
Pointed out in the community Slack, we don't show how to limit user workloads on control-plane only nodes.

### DIFF
--- a/docs/installation/server-roles.md
+++ b/docs/installation/server-roles.md
@@ -35,6 +35,15 @@ k3s-server-1   Ready    etcd                        5h39m   v1.20.4+k3s1
 k3s-server-2   Ready    control-plane,master        5h39m   v1.20.4+k3s1
 ```
 
+### Prevent workloads on control-plane nodes
+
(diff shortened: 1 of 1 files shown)


docs/installation/server-roles.md
```diff
@@ -35,6 +35,15 @@ k3s-server-1   Ready    etcd                        5h39m   v1.20.4+k3s1
 k3s-server-2   Ready    control-plane,master        5h39m   v1.20.4+k3s1
 ```
 
+### Prevent workloads on control-plane nodes
+
+By default, server nodes are schedulable and workloads can be launched on them. If you wish to have a dedicated control plane where no user workloads will run, use the `node-taint` parameter. Here is an example of adding a node taint to the configuration file:
```
brandond, line 40: We do, in the HA content. Should we have a dedicated section and link to it more prominently?

https://github.com/k3s-io/docs/blob/28542a1871c50e9d4a281bc0bc3f26d69a7a5823/docs/datastore/ha.md?plain=1#L46

dereknola: Yeah expanding this to include all node taints and linking to it in the HA page would be better.

briandowns (approved):

PR #209, opened 2023-11-28
Fix installation network-options table

Luigi600:
A small fix that corrects the presentation of the Markdown table, as well as formatting the table.

### DIFF
--- a/docs/installation/network-options.md
+++ b/docs/installation/network-options.md
@@ -19,15 +19,15 @@ This page describes K3s network configuration options, including configuration o
   You must ensure that WireGuard kernel modules are available on every node, both servers and agents, before attempting to use the WireGuard Flannel backend.
 
 
-|  CLI Flag and Value                  | Description                                                                                                                                                           |
-|--------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `--flannel-ipv6-masq`                | Apply masquerading rules to IPv6 traffic (default for IPv4). Only applies on dual-stack or IPv6-only clusters. Compatible with any Flannel backend other than `none`. |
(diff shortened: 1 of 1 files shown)


docs/installation/network-options.md
```diff
@@ -19,15 +19,15 @@ This page describes K3s network configuration options, including configuration o
...
+| `--flannel-backend=vxlan`            | Use VXLAN to encapsulate the packets. May require additional kernel modules on Raspberry Pi.                                                                          |
+| `--flannel-backend=host-gw`          | Use IP routes to pod subnets via node IPs. Requires direct layer 2 connectivity between all nodes in the cluster.                                                     |
+| `--flannel-backend=wireguard-native` | Use WireGuard to encapsulate and encrypt network traffic. May require additional kernel modules.                                                                      |
+| `--flannel-backend=ipsec`            | Use strongSwan IPSec via the `swanctl` binary to encrypt network traffic. (Deprecated; will be removed in v1.27.0)                                                    |
+| `--flannel-backend=none`             | Disable Flannel entirely.                                                                                                                                             |
```
brandond, line 30: Thanks for fixing the table formatting - but I don't think that it's necessary to pad these lines out with extra whitespace or hyphens to make the table line up in the raw view. Can you amend this to just fix the markdown?

brandond (changes requested):

docs/installation/network-options.md
Luigi600: Usually, such a formatting is generated when using code formatters such as Prettier. Basically, however, I have oriented myself to the following tables, which are part of the `main` branch:

* [Private Registry, Configs](https://github.com/k3s-io/docs/blob/c5fd3ff/docs/installation/private-registry.md?plain=1#L72-L85)
* [Requirements, "Inbound Rules for K3s Server Nodes" & "CPU and Memory"](https://github.com/k3s-io/docs/blob/c5fd3ff/docs/installation/requirements.md?plain=1#L136-L167)

If a change is nevertheless desirable, I will make it.

PR #211, opened 2023-11-30
Add reference for k3s-ansible repo in Quick-Start

### LINKED ISSUE #210: No reference to `k3s-ansible` repo in documentation
There is no reference to the [k3s-ansible](https://github.com/k3s-io/k3s-ansible) repository in the documentation.

This could be added in the [Quick-Start](https://docs.k3s.io/quick-start) page. This would allow easy bootstrap of multi-node cluster. Since the k3s-ansible repo is already part of the k3s-io system, i don't see how it would be a problem.

LombardiDaniel:
Closes #210 

Added reference for k3s-ansible repo in the [Quick-Start](https://docs.k3s.io/quick-start) section.

Screenshot showing changes on docusaurus:

Screenshot 2023-11-30 at 00 07 43


brandond:
cc @dereknola @cwayne18

dereknola:
I don't think we want to "recommend" the use of the k3s-ansible project. Pushing users to adopt Ansible is not our priority. The K3s docs largely remain about running and configuring K3s standalone. We have mentions of other projects such as K3d, only in the advanced section. Perhaps a better direction would be a section on "K3s ecosystem" projects, pointing things out such as K3d, K3sup, and the k3s-ansible repos.

sidebars.js
```diff
@@ -1,119 +1,119 @@
 module.exports = {
```
dereknola, line 1: Please revert the quotes changes to this file, standard Docusaurous convention is single quotes.

dereknola (changes requested):

LombardiDaniel:
didn't even realize i changed that file, my linter got it without me seeing it. All fixed now!

dereknola (approved):

docs/related-projects/related-projects.md
```diff
@@ -0,0 +1,16 @@
+---
```
brandond, line 1: I think you probably want to add this file to the sidebar list so that it shows up in the left-side TOC?
https://github.com/k3s-io/docs/blob/40e7e029624975cdbd3d5fca2dbb205bc6d0bbbb/sidebars.js#L1

brandond (changes requested):

brandond (approved):

PR #215, opened 2023-12-06
Improve wording of network requirements

wvell:


docs/installation/requirements.md
```diff
@@ -117,7 +117,7 @@ K3s performance depends on the performance of the database. To ensure optimal sp
 
 The K3s server needs port 6443 to be accessible by all nodes.
 
-The nodes need to be able to reach other nodes over UDP port 8472 when Flannel VXLAN is used or over UDP ports 51820 and 51821 (when using IPv6) when Flannel Wireguard backend is used. The node should not listen on any other port. K3s uses reverse tunneling such that the nodes make outbound connections to the server and all kubelet traffic runs through that tunnel. However, if you do not use Flannel and provide your own custom CNI, then the ports needed by Flannel are not needed by K3s.
+The nodes need to be able to reach other nodes over UDP port 8472 when using the Flannel VXLAN backend, and over UDP port 51820 (and 51821 if IPv6 is used) when using the Flannel WireGuard backend. The node should not listen on any other port. K3s uses reverse tunneling such that the nodes make outbound connections to the server and all kubelet traffic runs through that tunnel. However, if you do not use Flannel and provide your own custom CNI, then the ports needed by Flannel are not needed by K3s.
```
brandond, line 120:
```suggestion
The nodes need to be able to reach other nodes over UDP port 8472 when using the Flannel VXLAN backend, or over UDP port 51820 (and 51821 if IPv6 is used) when using the Flannel WireGuard backend. The node should not listen on any other port. K3s uses reverse tunneling such that the nodes make outbound connections to the server and all kubelet traffic runs through that tunnel. However, if you do not use Flannel and provide your own custom CNI, then the ports needed by Flannel are not needed by K3s.
```

brandond (changes requested):
nit - vxlan OR wg ports, not both.

brandond (approved):

dereknola (approved):

PR #216, opened 2023-12-07
Fix `sidebars.js` when accessing from `related-projects`

LombardiDaniel:
Just a fix `sidebars.js` when accessing from `related-projects`, the sidebar wasn't rendering when being accessed from that page.


brandond:
I almost called this out in review, but your screenshot showed that it was present in the sidebar so I figured it was working. Thanks for catching this!

PR #220, opened 2023-12-28
Fix traefik example helm chart

dereknola:
Addresses: https://github.com/k3s-io/docs/issues/218

### DIFF
--- a/docs/helm.md
+++ b/docs/helm.md
@@ -133,14 +133,11 @@ spec:
   valuesContent: |-
     image:
       name: traefik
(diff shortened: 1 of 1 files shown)


brandond (approved):

VestigeJ (commented):
What was the reason for dropping ssl?

VestigeJ (approved):

dereknola:
We aren't dropping, its no longer a valid configuration for traefik. They changed how permanent redirects are configured.

PR #243, opened 2024-02-15
Fix the `curl` command for preparing k3s images for Airgap

rishinair11:
Fixes https://github.com/k3s-io/docs/issues/244

This fixes an invalid URL issue that `curl` throws. 
```console
$ sudo curl -L -O /var/lib/rancher/k3s/agent/images/k3s-airgap-images-amd64.tar.zst https://github.com/k3s-io/k3s/releases/download/v1.29.1-rc2%2Bk3s1/k3s-airgap-images-amd64.tar.zst
curl: (3) URL using bad/illegal format or missing URL
Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
```

Since we are already passing the full path to the images.tar.zst, we can use `-o` instead of `-O`

### DIFF
--- a/docs/installation/airgap.md
+++ b/docs/installation/airgap.md
@@ -37,7 +37,7 @@ This method requires you to manually deploy the necessary images to each node, a
 2. Download the imagess archive to the agent's images directory, for example:
   ```bash
   sudo mkdir -p /var/lib/rancher/k3s/agent/images/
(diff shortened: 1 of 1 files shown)


brandond (approved):

brandond:
yeah, `-o` is the correct option, not `-O`. You shouldn't need to quote the URL but it doesn't hurt anything.

brandond (approved):
thanks for the contribution!

dereknola (approved):

rishinair11:
Thank you team. I appreciate the review!

PR #250, opened 2024-03-13
Add docs on wildcard registry

brandond:
Technically this has been supported by containerd for a long time, but we never documented or tested it. Ref:
* https://github.com/containerd/containerd/blob/release/1.7/docs/hosts.md#setup-default-mirror-for-all-registries
  > `Optionally the _default registry host namespace can be used as a fallback, if no other namespace matches.`
* https://github.com/containerd/containerd/blob/release/1.6/docs/cri/registry.md#configure-registry-endpoint
  > `* is configured, and gcr.io is not: endpoints for * + default endpoint https://gcr.io/v2.`

Also adds a mention of the two unsupported K3S_PTP env vars.

### DIFF
--- a/docs/installation/private-registry.md
+++ b/docs/installation/private-registry.md
@@ -130,6 +130,28 @@ The `auth` part consists of either username/password or authentication token:
 
 Below are basic examples of using private registries in different modes:
 
+### Wildcard Support
+
+:::info Version Gate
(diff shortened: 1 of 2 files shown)


briandowns (approved):

VestigeJ (approved):
LGTM

brandond:
Spell check failed on release notes, not from this PR.

PR #251, opened 2024-03-14
Update manual upgrade page with better order of operations

dereknola:
Addresses: https://github.com/k3s-io/docs/issues/178

~~Mentions that you have to stop the existing K3s service before running the installation script with a new version.~~

PR #254, opened 2024-03-18
Update datastore.md

swiedernix:
adjust PostgreSQL versions according these mentioned in https://github.com/k3s-io/kine/blob/master/scripts/test-run-postgres#L26-L31


brandond:
The certified versions are what our QA team has tested against, which isn't necessarily the same as what we run CI tests against.

cc @rancher-max

rancher-max (approved):
I'm fine with it since we have run those versions in CI! Our tests to certify dbs are fairly minimal as is anyway. Thank you!

brandond (approved):

dereknola:
Merging, codespell is fixed in https://github.com/k3s-io/docs/pull/251

PR #261, opened 2024-04-03
Add cli help anchor for Istio ref need

wilsonwu:
For Istio doc PR ref need: https://github.com/istio/istio.io/pull/14827


i18n/zh/docusaurus-plugin-content-docs/current/cli/server.md
```diff
@@ -180,7 +180,7 @@ weight: 1
 | `--flannel-backend` wireguard | N/A | 使用 `--flannel-backend=wireguard-native` |
 | `--flannel-backend` value=option1=value | N/A | 使用 `--flannel-conf` 指定带有后端配置的 Flannel 配置文件 |
 
-# K3s Server CLI 帮助
+# K3s Server CLI 帮助 {#k3s-server-cli-help}
```
dereknola, line 183:
```suggestion
## K3s Server CLI 帮助
```

dereknola (changes requested):
In the English version there are is "##" which is what generates the proper anchoring in docusaurous. That fix did not get applied to the zh translation for some reason.

brandond (approved):

PR #269, opened 2024-04-19
Remove unnecessary 'make-iptables-util-chains=true' from hardening guide

agterdenbos:
CIS [recommends](https://github.com/k3s-io/docs/blob/main/docs/security/self-assessment.md#427-ensure-that-the---make-iptables-util-chains-argument-is-set-to-true-automated) to *remove* the `--make-iptables-util-chains` argument from the `KUBELET_SYSTEM_PODS_ARGS` variable or set it as true. This parameter is already [true by default](https://github.com/kubernetes/kubernetes/blob/69b648a1d7074cbe004bf7adb3cdb17f01a4e9d8/pkg/kubelet/apis/config/v1beta1/defaults.go#L225). Therefore I suggest to remove this from the hardening guide.

### DIFF
--- a/docs/known-issues.md
+++ b/docs/known-issues.md
@@ -53,7 +53,6 @@ kube-controller-manager-arg:
   - 'use-service-account-credentials=true'
 kubelet-arg:
   - 'streaming-connection-idle-timeout=5m'
-  - 'make-iptables-util-chains=true'
 ```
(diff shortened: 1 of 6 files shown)


dereknola:
We can't remove this, as the CIS scanner itself needs to be modified to no longer check for `--make-iptables-util-chains`. We can likely remove this for K3s 1.30+ when cis-1.9 scans are published.

dereknola:
Closing in favor of https://github.com/k3s-io/docs/pull/328

PR #272, opened 2024-04-24
Add cis-1.8 assessment page

dereknola:
Also includes a script to generate markdown from kubebench json results.


cwayne18 (approved):

docs/security/cis-1.8.md
brandond, line 1: If the old page is self-assessment, do we want to call this one self-assessment-1.8 or something? Maybe make the old one self-assessment-1.23?

Also, how did CIS go from 1.23 -> 1.8!?

dereknola: Kubebench has a pretty good table on the version matching https://github.com/aquasecurity/kube-bench/blob/main/docs/platforms.md#cis-kubernetes-benchmark-support

I can rename the old page

brandond (approved):

brandond:
lgtm, once the translation stuff is fixed

PR #322, opened 2024-08-28
docs: added information to setup the tailscale VPN

JeremyTheocharis:
I spent quite some time in figuring it out until I realized I need to set ACL rules in tailscale.


brandond:
@manuelbuil ptal when you're back!

frzifus (approved):

manuelbuil (approved):

PR #330, opened 2024-10-04
IPv6 networking: mention --flannel-ipv6-masq in dual stack and single stack ipv6 section

recursivetree:
I run a cluster in dual stack mode using a cidr from the ULA range following the dual stack docs. Using [these](https://docs.k3s.io/networking/basic-network-options#dual-stack-ipv4--ipv6-networking) instructions, I noticed that pods use their pod IP for (cluster-external) outgoing connections, in this case meaning a ULA address, which obviously does not work. What I intended is that they use the host's address. After some googling, I found the `--flannel-ipv6-masq` option that does what I want. It is documented in the flannel section: https://docs.k3s.io/networking/basic-network-options#flannel-options. However, I think this should also be mentioned in the dual stack and single stack sections, as it is not given that people also read the flannel section.

This PR adds a mention of the need to add `--flannel-ipv6-masq` if using unroutable IPv6 ranges for a cluster.

I'd argue this use case is relatively common. For example, on many cloud platforms you just get a /64 or even a /128, so using something like ULAs is probablly a must.


brandond (approved):

brandond:
Thanks for the PR! Please sign-off your commit for DCO.

recursivetree:
Should be done, thanks. 

I've also opened an issue because the requirement for `Signed-off-by` in all commits is not mentioned in this repo: https://github.com/k3s-io/docs/issues/331

PR #334, opened 2024-10-11
Corrected namespaceSelector for network policy

larwef:
Corrected namespaceSelector for intra-namespace network policy example. The label "name" does not exist on the namespace resource kube-system, but the label "kubernetes.io/metadata.name" is present.

### DIFF
--- a/docs/security/hardening-guide.md
+++ b/docs/security/hardening-guide.md
@@ -380,7 +380,7 @@ spec:
     - from:
       - namespaceSelector:
           matchLabels:
-            name: kube-system
+            kubernetes.io/metadata.name: kube-system
 ```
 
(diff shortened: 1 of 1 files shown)


larwef:
Discovered this doing a fresh install on a Raspberry Pi 4 and going through the hardening guide. When trying to acquiring a Let's Encrypt certificate through a Traefik DNS requests timed out.

Not sure if v1.24 and Older is also affected by this. In case they should also be updated.

brandond (approved):

larwef:
Hi. I can force a new commit with the same changes. As far as i can see it's complaining about a dead link, which looks very much alive.

larwef:
Never mind. I got it now. Sorry, first time contributing.

I will fix.

PR #339, opened 2024-10-17
Update K3s Service Load Balancer docs

manuelbuil:
Explain a bit better the two uses cases that can happen and how the traffic flows for each of them

### DIFF
--- a/docs/networking/networking-services.md
+++ b/docs/networking/networking-services.md
@@ -58,15 +58,11 @@ Upstream Kubernetes allows Services of type LoadBalancer to be created, but does
 
 The ServiceLB controller watches Kubernetes [Services](https://kubernetes.io/docs/concepts/services-networking/service/) with the `spec.type` field set to `LoadBalancer`.
 
-For each LoadBalancer Service, a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) is created in the `kube-system` namespace. This DaemonSet in turn creates Pods with a `svc-` prefix, on each node. These Pods use iptables to forward traffic from the Pod's NodePort, to the Service's ClusterIP address and port.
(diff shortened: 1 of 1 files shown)


dereknola (approved):

docs/networking/networking-services.md
```diff
@@ -58,15 +58,12 @@ Upstream Kubernetes allows Services of type LoadBalancer to be created, but does
...
 
-If the ServiceLB Pod runs on a node that has an external IP configured, the node's external IP is populated into the Service's `status.loadBalancer.ingress` address list. Otherwise, the node's internal IP is used.
+When the ServiceLB Pod runs on a node that has an external IP configured, the node's external IP is populated into the Service's `status.loadBalancer.ingress` address list with `ipMode: VIP`. Otherwise, the node's internal IP is used.
 
-If multiple LoadBalancer Services are created, a separate DaemonSet is created for each Service.
+If the traffic to the external IP is natted, for example in public clouds when using the public IP of the node as external IP, the traffic is propagated into the ServiceLB Pod thanks to hostPort. The pod then uses iptables to forward traffic to the Service's ClusterIP address and port. If instead, the traffic is no natted, kube-proxy iptables chains `KUBE-EXT` will intercept the traffic and forward it to the Service's ClusterIP address and port.
```
brandond, line 65: A couple nits on language.

Do we want to perhaps skip mentioning the specific iptables chains used by kube-proxy, as this is kind of an implementation detail? The result is the same in ipvs mode, or when using a kube-proxy replacement like cilium, but they each do it a little differently.

```suggestion
If the traffic to the external IP is subject to [Network Address Translation (NAT)](https://en.wikipedia.org/wiki/Network_address_translation) - for example in public clouds when using the public IP of the node as external IP - the traffic is routed into the ServiceLB pod via the hostPort. The pod then uses iptables to forward traffic to the Service's ClusterIP address and port. If the traffic is not subject to NAT and instead arrives with the original destination address intact, the kube-proxy iptables `KUBE-EXT` chains will intercept the traffic and forward it to the Service's ClusterIP address and port.
```

brandond (changes requested):

docs/networking/networking-services.md
manuelbuil: Good point, I'll remove the `KUBE-EXT` part

PR #340, opened 2024-10-18
Add runtimes in reference

vitorsavian:
- This PR will add more info about the runtimes search in the new release that is coming from October

- I will wait for the releases and then merge the PR when got the approves

@dereknola do you think it is better to add a new place for the runtimes or we can add in the reference part? 

btw I feel like that `After October release` is not a very good thing to be wrote, should I add the versions?

### DIFF
--- a/docs/reference/env-variables.md
+++ b/docs/reference/env-variables.md
@@ -41,3 +41,24 @@ Environment variables which begin with `K3S_` will be preserved for the systemd 
 Setting `K3S_URL` without explicitly setting an exec command will default the command to "agent".
 
 When running the agent, `K3S_TOKEN` must also be set.
+
+:::info Version Gate
+Available as of October 2024 releases: v1.28.15+k3s1, v1.29.10+k3s1, v1.30.6+k3s1, v1.31.2+k3s1.
+:::
+
+K3s will now use `PATH` to find alternative container runtimes, in addition to checking the default paths used by the container runtime packages. In order to use this feature, you must modify the K3s service's PATH environment variable to add the directories containing the container runtime binaries.
+
+It's recommended that you modify one of this two environment files:
+
+- /etc/default/k3s # or k3s-agent
+- /etc/sysconfig/k3s # or k3s-agent
+
+This example will add the `PATH` in `/etc/default/k3s`:
+
+```bash
+echo PATH=$PATH >> /etc/default/k3s
+```
+
+:::warning
+`PATH` changes should be done with care to avoid placing untrusted binaries in the path of services that run as root.
+:::


dereknola:
Yeah change it to the "version gate" 
here is an example:
```
:::info Version Gate
Available as of September 2023 releases: v1.24.17+k3s1, v1.25.13+k3s1, v1.26.8+k3s1, v1.27.5+k3s1, v1.28.1+k3s1
:::
```

brandond (changes requested):

docs/reference/env-variables.md
```diff
@@ -40,4 +40,21 @@ Environment variables which begin with `K3S_` will be preserved for the systemd
...
+
+:::info Version Gate
+Available as of October 2024 releases: v1.28.15+k3s1, v1.29.10+k3s1, v1.30.6+k3s1, v1.31.2+k3s1.
+
+K3s will now use `PATH` to find runtimes, but it's necessary to add the `PATH` within the `k3s.service.env` or `k3s-agent.service.env` if you are using a agent.
+
+This example will add the `PATH` to `k3s.service.env` in the default directory:
+
+```bash
+echo PATH=$PATH >> /etc/systemd/system/k3s.service.env
```
brandond, line 53: This file is overwritten by the install script whenever it is run, we should not recommend that users modify it because changes to it may be lost. We should instead recommend one of the other two env files:
```
EnvironmentFile=-/etc/default/k3s # or k3s-agent
EnvironmentFile=-/etc/sysconfig/k3s # or k3s-agent
```

docs/reference/env-variables.md
```diff
@@ -40,4 +40,21 @@ Environment variables which begin with `K3S_` will be preserved for the systemd
...
+```
+
+:::warning
+
+`PATH` changes should be done with care to avoid security breaches or attacks.
```
brandond, line 58:
```suggestion
`PATH` changes should be done with care to avoid placing untrusted binaries in the path of services that run as root.
```

docs/reference/env-variables.md
```diff
@@ -40,4 +40,21 @@ Environment variables which begin with `K3S_` will be preserved for the systemd
...
 Setting `K3S_URL` without explicitly setting an exec command will default the command to "agent".
 
-When running the agent, `K3S_TOKEN` must also be set.
\ No newline at end of file
+When running the agent, `K3S_TOKEN` must also be set.
+
+:::info Version Gate
+Available as of October 2024 releases: v1.28.15+k3s1, v1.29.10+k3s1, v1.30.6+k3s1, v1.31.2+k3s1.
+
+K3s will now use `PATH` to find runtimes, but it's necessary to add the `PATH` within the `k3s.service.env` or `k3s-agent.service.env` if you are using a agent.
```
brandond, line 48:
```suggestion
K3s will now use `PATH` to find alternative container runtimes, in addition to checking the default paths used by the container runtime packages. In order to use this feature, you must modify the K3s service's PATH environment variable to add the directories containing the container runtime binaries.
```

brandond (changes requested):

brandond (approved):

docs/reference/env-variables.md
```diff
@@ -40,4 +40,26 @@ Environment variables which begin with `K3S_` will be preserved for the systemd
...
+
+```bash
+echo PATH=$PATH >> /etc/default/k3s
+```
+
+:::warning
+
+`PATH` changes should be done with care to avoid placing untrusted binaries in the path of services that run as root.
+
+:::
```
brandond, line 65:
```suggestion
:::warning
`PATH` changes should be done with care to avoid placing untrusted binaries in the path of services that run as root.
:::
```

docs/reference/env-variables.md
```diff
@@ -40,4 +40,26 @@ Environment variables which begin with `K3S_` will be preserved for the systemd
 
 Setting `K3S_URL` without explicitly setting an exec command will default the command to "agent".
 
-When running the agent, `K3S_TOKEN` must also be set.
\ No newline at end of file
+When running the agent, `K3S_TOKEN` must also be set.
+
+:::info Version Gate
+Available as of October 2024 releases: v1.28.15+k3s1, v1.29.10+k3s1, v1.30.6+k3s1, v1.31.2+k3s1.
```
brandond, line 46: End the INFO block
```suggestion
:::info Version Gate
Available as of October 2024 releases: v1.28.15+k3s1, v1.29.10+k3s1, v1.30.6+k3s1, v1.31.2+k3s1.
:::
```

brandond (changes requested):

brandond (approved):

dereknola (approved):

PR #351, opened 2024-11-01
Add the warning for externalTrafficPolicy=Local

manuelbuil:
Describe how we could preserve the client source IP

### DIFF
--- a/docs/networking/networking-services.md
+++ b/docs/networking/networking-services.md
@@ -68,6 +68,10 @@ If the traffic to the external IP is subject to [Network Address Translation (NA
 
 Create a [Service of type LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) in K3s.
 
+:::warning Known Issue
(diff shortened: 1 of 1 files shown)


docs/networking/networking-services.md
```diff
@@ -68,6 +68,10 @@ If the traffic to the external IP is subject to [Network Address Translation (NA
 
 Create a [Service of type LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) in K3s.
 
+:::warning Known Issue
+If external traffic reaches the node using a NAT (e.g. in public clouds) and you require `externalTrafficPolicy=local` for purposes such as client source IP preservation, please do not define the k3s config `node-external-ip` for any of the nodes, as that will not work correctly
```
VestigeJ, line 72: nit: punctuation period at the end

without setting the node-external-ip people will have to set the tls-san correctly to be able to join extra nodes correct?

brandond: Having node-external-ip would be necessary for flannel to work properly, if all the nodes aren't on the same private network. So I guess this would be a tradeoff.

manuelbuil: exactly. People should use the internal-ip to join the cluster

VestigeJ (approved):

PR #357, opened 2024-11-13
Add useful information for tokens

manuelbuil:
Added some extra information which could be useful for other users

### DIFF
--- a/docs/cli/token.md
+++ b/docs/cli/token.md
@@ -4,7 +4,7 @@ title: token
 
 # k3s token
 
-K3s uses tokens to secure the node join process. Tokens authenticate the cluster to the joining node, and the node to the cluster.
+K3s uses tokens to secure the node join process and to encrypt confidential information that is persisted to the datastore. Tokens authenticate the cluster to the joining node, and the node to the cluster.
 
 ## Token Format
 
@@ -51,9 +51,9 @@ Bootstrap | `n/a`           | `n/a`
 
 If no token is provided when starting the first server in the cluster, one is created with a random password. The server token is always written to `/var/lib/rancher/k3s/server/token`, in secure format.
 
-The server token can be used to join both server and agent nodes to the cluster. It cannot be changed once the cluster has been created, and anyone with access to the server token essentially has full administrator access to the cluster. This token should be guarded carefully.
+The server token can be used to join both server and agent nodes to the cluster. Anyone with access to the server token essentially has full administrator access to the cluster. This token should be guarded carefully.
 
-The server token is also used as the [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2) passphrase for the key used to encrypt confidential information that is persisted to the datastore, such as the secrets-encryption configuration, wireguard keys, and private keys for cluster CA certificates and service-account tokens. For this reason, the token must be backed up alongside the cluster datastore itself.
+The server token is also used as the [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2) passphrase to encrypt confidential information that is persisted to the datastore known as bootstrap data. Bootstrap data is essential to set up new server nodes or restore from a snapshot. For this reason, the token must be backed up alongside the cluster datastore itself.
 
 :::warning
 Unless custom CA certificates are in use, only the short (password-only) token format can be used when starting the first server in the cluster. This is because the cluster CA hash cannot be known until after the server has generated the self-signed cluster CA certificates.
@@ -70,11 +70,7 @@ The agent token is written to `/var/lib/rancher/k3s/server/agent-token`, in secu
 
 ### Bootstrap
 
-:::info Version Gate
-Support for the `k3s token` command and the ability to join nodes with bootstrap tokens is available starting with the 2023-02 releases (v1.26.2+k3s1, v1.25.7+k3s1, v1.24.11+k3s1, v1.23.17+k3s1).
-:::
-
-K3s supports dynamically generated, automatically expiring agent bootstrap tokens. Bootstrap tokens can only be used to join agents.
+K3s supports dynamically generated, automatically expiring agent [bootstrap tokens](https://kubernetes.io/docs/reference/access-authn-authz/bootstrap-tokens/).
 
 ## k3s token
 
@@ -149,7 +145,7 @@ Flag | Description
 Available as of the October 2023 releases (v1.28.2+k3s1, v1.27.7+k3s1, v1.26.10+k3s1, v1.25.15+k3s1).
 :::
 
-Rotate original server token with a new bootstrap token. After running this command, all servers and any agents that originally joined with the old token must be restarted with the new token.
+Rotate original server token with a new server token. After running this command, all servers and any agents that originally joined with the old token must be restarted with the new token.
 
 If you do not specify a new token, one will be generated for you.
 
@@ -161,3 +157,6 @@ If you do not specify a new token, one will be generated for you.
 `--token` value      | Existing token used to join a server or agent to a cluster [$K3S_TOKEN]                   
 `--new-token` value  | New token that replaces existing token   
 
+:::warning
+Snapshots taken before the rotation will require the old server token when restoring the cluster
+:::


docs/cli/token.md
```diff
@@ -161,3 +161,10 @@ If you do not specify a new token, one will be generated for you.
 `--token` value      | Existing token used to join a server or agent to a cluster [$K3S_TOKEN]                   
 `--new-token` value  | New token that replaces existing token   
 
+:::warning
+Data encrypted before the rotation will need the old server token to decrypt
+:::
+
+:::warning
+The new server token is not written in `/var/lib/rancher/k3s/server/token`
```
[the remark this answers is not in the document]
dereknola: Yeah this is not the intended result. The server should be writing to that file https://github.com/k3s-io/k3s/blob/master/pkg/server/token.go#L89-L92

docs/cli/token.md
```diff
@@ -161,3 +161,10 @@ If you do not specify a new token, one will be generated for you.
 `--token` value      | Existing token used to join a server or agent to a cluster [$K3S_TOKEN]                   
 `--new-token` value  | New token that replaces existing token   
 
+:::warning
+Data encrypted before the rotation will need the old server token to decrypt
```
dereknola, line 165: We attempt to migrate all data that depends on those keys with https://github.com/k3s-io/k3s/blob/master/pkg/cluster/storage.go#L304. What data is not being reencrypted?

docs/cli/token.md
```diff
@@ -51,7 +51,7 @@ Bootstrap | `n/a`           | `n/a`
 
 If no token is provided when starting the first server in the cluster, one is created with a random password. The server token is always written to `/var/lib/rancher/k3s/server/token`, in secure format.
 
-The server token can be used to join both server and agent nodes to the cluster. It cannot be changed once the cluster has been created, and anyone with access to the server token essentially has full administrator access to the cluster. This token should be guarded carefully.
+The server token can be used to join both server and agent nodes to the cluster. Unless manually rotated with [`k3s token rotate`](#k3s-token-rotate), the server token does not change and anyone with access to the server token essentially has full administrator access to the cluster. This token should be guarded carefully.
```
brandond, line 54: I don't think this belongs here. The fact that it can be rotated does not alter the fact that anyone who has access to it has full admin access.  If you want to call out that it can/should be rotated if compromised, do so in a separate paragraph or sentence.

docs/cli/token.md
```diff
@@ -74,7 +74,7 @@ The agent token is written to `/var/lib/rancher/k3s/server/agent-token`, in secu
 Support for the `k3s token` command and the ability to join nodes with bootstrap tokens is available starting with the 2023-02 releases (v1.26.2+k3s1, v1.25.7+k3s1, v1.24.11+k3s1, v1.23.17+k3s1).
 :::
 
-K3s supports dynamically generated, automatically expiring agent bootstrap tokens. Bootstrap tokens can only be used to join agents.
+K3s supports dynamically generated, automatically expiring agent [bootstrap tokens](https://kubernetes.io/docs/reference/access-authn-authz/bootstrap-tokens/). Bootstrap tokens can only be used to join agents and are consumed like the agent tokens with the config `--agent-token` or `K3S_AGENT_TOKEN`.
```
brandond, line 77:
```suggestion
K3s supports dynamically generated, automatically expiring agent [bootstrap tokens](https://kubernetes.io/docs/reference/access-authn-authz/bootstrap-tokens/). Bootstrap tokens can only be used to join agents and are consumed like the agent tokens with the config `--agent-token` or `K3S_AGENT_TOKEN`.
```

This is not correct. You do not need to set --agent-token to use bootstrap tokens. They are created using the CLI  or kubectl.

docs/cli/token.md
```diff
@@ -149,7 +149,7 @@ Flag | Description
 Available as of the October 2023 releases (v1.28.2+k3s1, v1.27.7+k3s1, v1.26.10+k3s1, v1.25.15+k3s1).
 :::
 
-Rotate original server token with a new bootstrap token. After running this command, all servers and any agents that originally joined with the old token must be restarted with the new token.
+Rotate original server token with a new bootstrap token. After running this command, all servers and any agents that originally joined with the old token must be restarted with the new token. This bootstrap token never expires.
```
brandond, line 152: Ugh our reuse of the term "bootstrap" sucks. This is the server token used to encrypt the bootstrap data, it is NOT an "agent bootstrap token" and I think we should not use the term "bootstrap token" to refer to anything other than an agent bootstrap token. This should just say "server token".

docs/cli/token.md
brandond: Yeah I don't think this is accurate. Changing the server token does NOT change the encryption config, so the same key is used to encrypt/decrypt the secrets.

docs/cli/token.md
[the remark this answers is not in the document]
brandond: 👍🏻 on above concerns

brandond (changes requested):

docs/cli/token.md
manuelbuil: My intention was to change the sentence: `It cannot be changed once the cluster has been created`. This is not accurate because it surely can be changed with `k3s token rotate`. I can try to find some better wording

docs/cli/token.md
manuelbuil: I wanted to express that, to consume the already created bootstrap token, you must use `--agent-token` or `K3S_AGENT_TOKEN`, or is there any other way? We were a bit confused by the matrix https://docs.k3s.io/cli/token#token-types as it was not clear how to consume the bootstrap tokens.

docs/cli/token.md
manuelbuil: Ok, let me try to rewrite it a bit better

docs/cli/token.md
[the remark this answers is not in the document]
manuelbuil: Sorry, I thought I had tested it but now I realize this is wrong :(. Thanks for catching that up

docs/cli/token.md
manuelbuil: Ok, I got confused by a comment. The only encrypted object which will need the old server token are old snapshots then, right? I wanted to add this warning because when executing this command, you get a warning with something like `keep the old token` but perhaps it is too late once you read that warning? The token gets replaced a few seconds after that

docs/cli/token.md
brandond: On agents, you just pass the token via --token, corresponding config key, or the K3S_TOKEN env var.

--agent-token is set on servers to generate a token that can only be used to join agents. Before bootstrap token support, this was the only way to get an unprivileged token that could not be used to join servers.

docs/cli/token.md
brandond: Lets just take that out, and start the second sentence with "Anyone with access to the server token...". We can get into rotation later.

docs/cli/token.md
brandond: The snapshots themselves are not encrypted either. Any snapshots taken with the old token value, will still have the bootstrap data IN the snapshot encrypted with the old token, so you'd need to use the old token when restoring it.

manuelbuil: Thanks Brad, for some reason I was misunderstanding this. I thought the snapshot was also encrypted :facepalm:

brandond: Its kinda complicated. Everything in this struct:
https://github.com/k3s-io/k3s/blob/cd4ddedbc9782cbe9b5dcc411df2addae7b2f3b4/pkg/daemons/config/types.go#L291-L307

Is read from disk is stuffed into a map, where the key is the field name and the value is the content of file at that path on disk along with a modification timestamp. This is then serialized to JSON and encrypted with the key.

You can find the code for translating from `ControlRuntimeBootstrap` to `map[string]File` here:
https://github.com/k3s-io/k3s/blob/cd4ddedbc9782cbe9b5dcc411df2addae7b2f3b4/pkg/bootstrap/bootstrap.go#L23-L54

manuelbuil: Thanks Brad. Since the struct https://github.com/k3s-io/k3s/blob/cd4ddedbc9782cbe9b5dcc411df2addae7b2f3b4/pkg/daemons/config/types.go#L291-L307 might keep changing in the future, I think it is better if we remove what comes after `such as` to avoid it getting outdated again. I rewrote the paragraph a bit

dereknola (approved):

brandond (approved):

VestigeJ (approved):

PR #366, opened 2024-12-19
Replace binary with process

manuelbuil:
Basically sync with https://github.com/rancher/rke2-docs/pull/297. There is just one thing

### DIFF
--- a/docs/upgrades/manual.md
+++ b/docs/upgrades/manual.md
@@ -69,5 +69,5 @@ To upgrade K3s manually, you can download the desired version of the K3s binary 
 
 1. Download the desired version of the K3s binary from [releases](https://github.com/k3s-io/k3s/releases)
 2. Copy the downloaded binary to `/usr/local/bin/k3s` (or your desired location)
-3. Stop the old k3s binary
-4. Launch the new k3s binary
+3. Stop the old k3s process
+4. Launch the new k3s process


brandond (changes requested):

docs/upgrades/manual.md
```diff
@@ -69,5 +69,5 @@ To upgrade K3s manually, you can download the desired version of the K3s binary
 
 1. Download the desired version of the K3s binary from [releases](https://github.com/k3s-io/k3s/releases)
 2. Copy the downloaded binary to `/usr/local/bin/k3s` (or your desired location)
-3. Stop the old k3s binary
-4. Launch the new k3s binary
\ No newline at end of file
+3. Stop the old k3s process
+4. Launch the new k3s process
```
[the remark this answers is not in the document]
manuelbuil: Yeah, definitely makes more sense

brandond (approved):

docs/upgrades/manual.md
```diff
@@ -69,5 +69,4 @@ To upgrade K3s manually, you can download the desired version of the K3s binary
 
 1. Download the desired version of the K3s binary from [releases](https://github.com/k3s-io/k3s/releases)
 2. Copy the downloaded binary to `/usr/local/bin/k3s` (or your desired location)
-3. Stop the old k3s binary
-4. Launch the new k3s binary
\ No newline at end of file
+3. Restart the k3s or k3s-agent service
```
dereknola, line 72:
```suggestion
3. Restart the k3s or k3s-agent service (systemd/openrc)
    OR
    Restart the k3s process (binary)
```

RKE2 does not operate as a standalone binary, but K3s very much can. So we need to to include both options. We just talked about the binary in the previous steps.

dereknola (changes requested):

docs/upgrades/manual.md
brandond: They both function the same when run from a terminal. I really hope that noone is running k3s from a shell. Honestly I would like to revise the rest of the docs to no longer suggest just running `k3s server` to start a node.

dereknola: I mean.... isn't that what our k3s docker image does, run K3s in a shell, not a service?

brandond: Yes, but also you can't upgrade the k3s image by replacing the binary and killing the process.

brandond (approved):

dereknola (approved):

PR #371, opened 2025-01-18
[Translation] pt-BR

italopenaforte:
Hi guys

lately have been studying to migrating my homelab to use Kubernetes and i'm start using k3s. While i'm studying k3s i will add translation for brazilian portugues


dereknola:
After accepting several translations as one time PRs, we are no longer accepting additional languages, as people do not maintain the languages they submit. This has led to more work, as we have to maintain languages we do not natively speak/read. More trouble than it is worth.

PR #376, opened 2025-01-29
Remove duplicate info about tokens and certificates

manuelbuil:
In the Advanced Options section there is information about certificates and tokens which is also in the specific sections for tokens and certificates. I guess that information started in the Advanced Options when there wasn't a specific version for them and then we forgot to remove it

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -4,26 +4,6 @@ title: "Advanced Options / Configuration"
 
 This section contains advanced information describing the different ways you can run and manage K3s, as well as steps necessary to prepare the host OS for K3s use.
 
-## Certificate Management
-
-### Certificate Authority Certificates
-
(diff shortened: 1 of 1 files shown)


brandond (changes requested):
I think its a good idea to have brief info here, and then direct people to the specific CLI pages for details. Maybe we just fix the outdated info in the Token Management section?

manuelbuil:
> I think its a good idea to have brief info here, and then direct people to the specific CLI pages for details. Maybe we just fix the outdated info in the Token Management section?

In my opinion, this "advanced options" section became some kind of junk drawer where we put a lot of unconnected stuff. I think we should try to define a bit more what content makes sense here and what content does not. My interpretation was that, as of today, the information we were adding, was important information which did not fit in other sections. Therefore, as certificates and tokens have now their own section, it does not make sense to continue having them here.

However, if we believe we should have a brief info on stuff related to CLI tools in the "advanced options" section, we should also add some lines describing secrets-encrypt and etcd-snapshot, right? Let's talk in SU and see what other think too

PR #381, opened 2025-02-03
manual upgrades: clarify that a drain is not required

kyrofa:
In Kubernetes, pods are independent of containerd/kubelet, and thus restarting k3s without draining first is a safe thing to do. This is not immediately clear to folks new to hosting their own cluster, so add a reassuring note.

### DIFF
--- a/docs/upgrades/manual.md
+++ b/docs/upgrades/manual.md
@@ -42,6 +42,10 @@ Running the install script will:
 2. Update the systemd unit or openrc init script to reflect the args passed to the install script
 3. Restart the k3s service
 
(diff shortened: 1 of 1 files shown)


brandond:
thanks for the contribution! looks like this needs a rebase.

farazkhawaja:
@brandond on a second thought, should we go this route?https://github.com/kubernetes/kubernetes/issues/129385#issuecomment-2574345344
Upgrades to 1.31 leads to container restarts, i see upstream k8s maintainers recommend cordon+drain for this situation.

PR #402, opened 2025-04-08
Update misc docs pages

brandond:
Update helm and containerd template docs to sync with recent changes to helm-controller and containerd 2.0. Also address some nits on token/proxy/ccm.


VestigeJ (commented):
can't approve but LGTM

PR #407, opened 2025-04-16
Remove outdated storage information

manuelbuil:
Kubernetes is not maintaining in-tree volume plugins anymore, we should remove that information from the docs


brandond (approved):
💯

dereknola (approved):

PR #415, opened 2025-05-20
Add 1.33 release notes, new sidebar collapse for older releases

dereknola:
image

Note that this does change the link to older release notes. (/release-notes -> /release-notes-old)


brandond:
Can we add the old pages to the redirects, just to keep them working under their original paths?
https://github.com/k3s-io/docs/blob/30412a10e46cf834a085314fc2c168393ab4aa5b/docusaurus.config.js#L129-L131

dereknola:
I was worried it would become unwieldy in the future of continually adding redirects to this list. But we can for now.

dereknola:
> Can we add the old pages to the redirects, just to keep them working under their original paths?

This is not an option due to https://github.com/facebook/docusaurus/issues/9505

brandond:
welp

brandond (approved):

PR #453, opened 2025-07-30
Update automated.md

farazkhawaja:
address a small confusion where a user might think that their node will always remain cordoned following a failed downgrade plan. Only users with `cordon:true` will undergo such situation


brandond:
Can you rebase to main? We just pushed an update that fixes some external links that have changed.

farazkhawaja:
@brandond workflow requires re-rerun

farazkhawaja:
@brandond would you know why is the check failing? its synced with main

brandond:
don't worry about it, something external is just being flakey. it'll pass eventually.

PR #455, opened 2025-08-01
Change rke2-multus configuration when version >= v4.2.202

thomasferrandiz:


docs/networking/multus-ipams.md
```diff
@@ -46,6 +46,8 @@ spec:
         confDir: /var/lib/rancher/k3s/agent/etc/cni/net.d
         binDir: /var/lib/rancher/k3s/data/cni/
         kubeconfig: /var/lib/rancher/k3s/agent/etc/cni/net.d/multus.d/multus.kubeconfig
+        # Uncomment the following line when using rke2-multus >= v4.2.202
```
brandond, line 49: I'm kinda assuming anyone reading the docs will be installing this for the first time, and will get a version of the chart that requires this. Maybe just include it by default? Does anything bad happen if it's set on older chart versions?

brandond (approved):

PR #463, opened 2025-08-14
Clarify steps and add drop-down menu

manuelbuil:
Clarify some steps in the manually deploy images section and add a drop-down menu:
image

and

image


brandond (changes requested):
We should add additional information to the summaries. With the sections collapsed by default, users will miss critical information that might tell them whether or not they need to perform these steps.

PR #490, opened 2025-10-20
feat(i18n): translate installation instructions gateway page to Chinese

kamaslau:


kamaslau:
Updated with the correct file path: https://github.com/k3s-io/docs/pull/492

PR #496, opened 2025-10-21
Consildate several pages under Add-Ons section

dereknola:
## Changes
Moved `helm`, `Import Images`, and `Storage` pages under an `Add-Ons` section, which encompasses additional functionality that is bundled into K3s, but comes from other software packages (helm, spegel, longhorn etc..).

## Future Work
Move information about GPUs and Alternative Runtimes under this section. 

image


dereknola:
We went with [Add-Ons for the RKE2 docs](https://github.com/rancher/rke2-docs/pull/423) as well, so I was matching naming.  `Packaged Components` makes me think of the stuff like crictl, busybox, and all the other binaries that K3s is shipped with, not the Feature Functionality of it. 

`Integrations` is not the right word because the other software producers aren't working "with K3s" in some way, we are co-opting the software. 

I don't have a good word for "A feature of K3s that exists because of external software packages, but require user intervention to operate (e.g. place manifests, deploy PVCs, place image.tar.gz files)"

How about "Extensions"?

brandond (approved):
I guess if we're just creating a new category without changing any of the wording around addons/packaged components/bundled binaries... that's probably not too confusing.

PR #567, opened 2026-05-29
Document the imports/drop-in approach for configuring containerd (#566)

### LINKED ISSUE #566: Should `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d` be documented?
I recently needed to amend the K3s' containerd configuration, specifically I wanted to add
```
[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]
  cgroup_writable = true
```
to enable support for running systemd-based containers with `hostUsers: false`, for https://github.com/freeipa/freeipa-container/issues/725.

I followed https://docs.k3s.io/advanced#configuring-containerd and https://docs.k3s.io/advanced#base-template and created `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.tmpl` with content
```
{{ template "base" . }}
[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]
  cgroup_writable = true
```

I then saw the containerd fail to start with `/var/lib/rancher/k3s/agent/containerd/containerd.log` having a stream of
```
containerd: failed to unmarshal TOML: toml: table runc already exists
containerd: failed to unmarshal TOML: toml: table runc already exists
containerd: failed to unmarshal TOML: toml: table runc already exists
```
messages.

I checked `/var/lib/rancher/k3s/agent/etc/containerd/config.toml` and it indeed already has section
```
[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]
  runtime_type = "io.containerd.runc.v2"
```
there, so there were two `[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]`s there in the resulting `config.toml`. Looking at https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go, there is currently no way to customize that default section.

For a while it seemed I would have to copy the whole template from https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go (instead of using the recommended `{{ template "base" . }}` approach) but then I noticed
```
imports = ["/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/*.toml"]
```
at the top of `/var/lib/rancher/k3s/agent/etc/containerd/config.toml`.

Being brave I did not even try to read what it does, I just created `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/cgroups.toml` with
```
version = 3
[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]
  cgroup_writable = true
```
... and it seems to do the trick, containerd starts and systemd-based containers with `hostUsers: false` also work fine.

I consider drop-in files to be easier than fiddling with templates, so I wonder if the approach which seems to work for me is supported and will work in the future as well, and if yes, whether it should be documented next to (potentially above of) the current https://docs.k3s.io/advanced#base-template section.

adelton:
A proposal for documenting the `imports` / drop-in containerd configuration, added via https://github.com/k3s-io/k3s/pull/13680.

I ordered the sections to start with the drop-in approach as I feel that's what people might want to try first, going forward. I minimized the information about pre-containerd-2 setups as I assume folks that have those and had a need for specific containerd configuration have already their systems set up.

The Base template section got quite long for my taste. I wanted to document the pitfalls people might hit with that approach (the issue I hit) but maybe if noone complained so far, that hand-holding is not necessary? Alternatively it could mean noone really uses the base template approach?

Comments welcome.

Resolves https://github.com/k3s-io/docs/issues/566.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -168,28 +168,36 @@ sudo etcdctl version \
 
 ## Configuring containerd
 
-:::info Version Gate
-K3s includes containerd 2.0 as of the February 2025 releases: v1.31.6+k3s1 and v1.32.2+k3s1.  
-Be aware that containerd 2.0 prefers config version 3, while containerd 1.7 prefers config version 2.
-:::
+K3s will generate a configuration file for containerd at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml` using values specific to the current cluster and node configuration.
 
-K3s will generate a configuration file for containerd at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml`, using values specific to the current cluster and node configuration.
+Starting with February 2025 releases, K3s includes containerd 2+ and the configuration file declares and uses the `version = 3` format.
+Starting with April 2026 K3s releases, the default generated `config.toml` includes `imports` directive with the `config-v3.toml.d/*.toml` path.
 
-For advanced customization, you can create a containerd config template in the same directory:
-* For containerd 2.0, place a version 3 configuration template in `config-v3.toml.tmpl`  
-  See the [containerd 2.0 documentation](https://github.com/containerd/containerd/blob/release/2.0/docs/cri/config.md) for more information.
-* For containerd 1.7 and earlier, place a version 2 configuration template in `config.toml.tmpl`  
-  See the [containerd 1.7 documentation](https://github.com/containerd/containerd/blob/release/1.7/docs/cri/config.md) for more information.
+So with recent K3s versions, the simplest way to tweak specific configuration values while keeping the K3s-provided defaults is to create the `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d` directory and put a drop-in file with just the values you wish to change there:
 
-Containerd 2.0 is backwards compatible with prior config versions, and k3s will continue to render legacy version 2 configuration from `config.toml.tmpl` if `config-v3.toml.tmpl` is not found.
+```toml title="/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/cgroups.toml"
+version = 3
+[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]
+  cgroup_writable = true
+```
 
+To quote the [containerd-config.toml man page(5)](https://containerd.io/docs/2.3/man/containerd-config.toml.5/):
+
+> Imported files will overwrite simple fields like `int` or `string` (if not empty) and will append `array` and `map` fields.
+
+If you wish to provide a full containerd configuration file instead of the default one, you can create a containerd config template `config-v3.toml.tmpl` in `/var/lib/rancher/k3s/agent/etc/containerd`.
 The template file is rendered into the containerd config using the [`text/template`](https://pkg.go.dev/text/template) library.
-See `ContainerdConfigTemplateV3` and `ContainerdConfigTemplate` in [`templates.go`](https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go) for the default template content.
+Search for `ContainerdConfigTemplateV3` in [`templates.go`](https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go) for the default template content.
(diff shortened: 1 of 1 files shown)


adelton:
Cc: @fidencio.

docs/advanced.md
```diff
@@ -168,28 +168,36 @@ sudo etcdctl version \
...
-K3s includes containerd 2.0 as of the February 2025 releases: v1.31.6+k3s1 and v1.32.2+k3s1.  
-Be aware that containerd 2.0 prefers config version 3, while containerd 1.7 prefers config version 2.
-:::
+K3s will generate a configuration file for containerd at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml` using values specific to the current cluster and node configuration.
+
+Starting with February 2025 releases, K3s includes containerd 2+ and the configuration file declares and uses the `version = 3` format.
+Starting with April 2026 K3s releases, the default generated `config.toml` includes `imports` directive with the `config-v3.toml.d/*.toml` path.
 
-K3s will generate a configuration file for containerd at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml`, using values specific to the current cluster and node configuration.
+So with recent K3s versions, the simplest way to tweak specific configuration values while keeping the K3s-provided defaults is to create the `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d` directory and put a drop-in file with just the values you wish to change there:
```
brandond, line 178: I think this is a bit excessively informal / conversational. Can you retain the existing style?

docs/advanced.md
```diff
@@ -168,28 +168,36 @@ sudo etcdctl version \
 
 ## Configuring containerd
 
-:::info Version Gate
-K3s includes containerd 2.0 as of the February 2025 releases: v1.31.6+k3s1 and v1.32.2+k3s1.  
-Be aware that containerd 2.0 prefers config version 3, while containerd 1.7 prefers config version 2.
-:::
+K3s will generate a configuration file for containerd at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml` using values specific to the current cluster and node configuration.
+
+Starting with February 2025 releases, K3s includes containerd 2+ and the configuration file declares and uses the `version = 3` format.
```
brandond, line 174: Please use the `:::info Version Gate` infobox format when mentioning what releases support a feature.

docs/advanced.md
```diff
@@ -168,28 +168,36 @@ sudo etcdctl version \
...
 
+If you wish to provide a full containerd configuration file instead of the default one, you can create a containerd config template `config-v3.toml.tmpl` in `/var/lib/rancher/k3s/agent/etc/containerd`.
 The template file is rendered into the containerd config using the [`text/template`](https://pkg.go.dev/text/template) library.
-See `ContainerdConfigTemplateV3` and `ContainerdConfigTemplate` in [`templates.go`](https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go) for the default template content.
+Search for `ContainerdConfigTemplateV3` in [`templates.go`](https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go) for the default template content.
 The template is executed with a [`ContainerdConfig`](https://github.com/k3s-io/k3s/blob/main/pkg/agent/templates/templates.go#L22-L33) struct as its dot value (data argument).
+For best results and to get your future cluster and node configuration changes reflected properly, do NOT simply copy a prerendered `config.toml` with your changes into the template location, provide the template file with template markup.
```
brandond: use the existing language for this, even if you are going to move it elsewhere.

docs/advanced.md
adelton: Non-native speaker here -- could you be more specific about what aspects you dislike? The file has a bunch of "you want" / "you can" there ...

docs/advanced.md
adelton: Added back in b24738232646a79a73c0b12aece904f7dc92e1f7 (to be squashed eventually).

docs/advanced.md
adelton: Done in e3f903b8a162f34493df56e3746ff4caf58d887f (to be squashed eventually).

The problem is that the original paragraph refers to the base template so should there be a "see below" note there?

I still feel that a reason for using the template approach would be nice to be spelled out.

PR #581, opened 2026-07-22
Drop EOL Version Gates

manuelbuil:
There are a few Version Gates which are pointing at releases that are all EOLed. This PR removes them

### DIFF
--- a/docs/cli/etcd-snapshot.md
+++ b/docs/cli/etcd-snapshot.md
@@ -106,6 +106,11 @@ K3s supports replicating etcd snapshots to and restoring etcd snapshots from S3-
 | `--etcd-s3-timeout` | S3 timeout (default: `5m0s`) |
 | `--etcd-s3-config-secret` | Name of secret in the kube-system namespace used to configure S3, if etcd-s3 is enabled and no other etcd-s3 options are set |
 
+:::note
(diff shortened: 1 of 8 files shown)


docs/cli/etcd-snapshot.md
```diff
@@ -105,6 +105,7 @@ K3s supports replicating etcd snapshots to and restoring etcd snapshots from S3-
 | `--etcd-s3-insecure` | Disables S3 over HTTPS |
 | `--etcd-s3-timeout` | S3 timeout (default: `5m0s`) |
 | `--etcd-s3-config-secret` | Name of secret in the kube-system namespace used to configure S3, if etcd-s3 is enabled and no other etcd-s3 options are set |
+| `--etcd-s3-retention` | Number of snapshots in S3 to retain (default: `5`) |
```
brandond: I'm not sure how to express this exactly in the CLI help, but the value of  `--etcd-snapshot-retention` also sets `--etcd-s3-retention`, if it is not explicitly set in the config.

That is what we were expressing with the "It has the same default value as the local snapshot retention" comment.

brandond (approved):
lgtm with one nit

docs/cli/etcd-snapshot.md
manuelbuil: let's add a `note`. Besides, I realized, `--etcd-s3-retention` is already in this matrix

brandond (approved):

PR #589, opened 2026-08-12
Add example for Docker Hub login to registries.yaml

dereknola:
Because even K3s devs forget things. 

### DIFF
--- a/docs/installation/private-registry.md
+++ b/docs/installation/private-registry.md
@@ -51,7 +51,7 @@ configs:
     auth:
       username: <BASIC AUTH USERNAME>
       password: <BASIC AUTH PASSWORD>
-      token: <BEARER TOKEN>
+      auth: <BASE64 USERNAME:PASSWORD>
     tls:
       ca_file: <PATH TO SERVER CA>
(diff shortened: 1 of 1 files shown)


docs/installation/private-registry.md
```diff
@@ -47,7 +51,7 @@ configs:
     auth:
       username: <BASIC AUTH USERNAME>
       password: <BASIC AUTH PASSWORD>
-      token: <BEARER TOKEN>
+      auth: <BASE64 USERNAME:PASSWORD>
```
brandond, line 54: Did we confirm that this is the **only** thing that could possibly go here? I can't remember how containerd passes this through. Are the username and password fields just conveniences to save on having to join/b64encode the values and stick them in this field?

brandond (approved):

docs/installation/private-registry.md
dereknola: Yes thats what it is. 
https://github.com/containerd/containerd/blob/main/core/remotes/docker/authorizer.go#L271

The docs are super unclear on this.

